PR #10, opened 2020-11-28
Decomposition and normalisation

pevnak:
Hi,

these are implementations of those two functions we have discussed in the Setfield thread.
They works as
```
julia> optic = (@optic _.c) ∘ (@optic _.a.b)
(@optic _.c) ∘ (@optic _.b) ∘ (@optic _.a)

julia> optic.inner
(@optic _.b) ∘ (@optic _.a)

julia> optic.outer
(@optic _.a)

julia> normalise(optic).inner
(@optic _.a)

julia> normalise(optic).outer
(@optic _.c) ∘ (@optic _.b)
```

and
```
julia> decompose(@optic _.a.b.c.d)
((@optic _.d), (@optic _.c), (@optic _.b), (@optic _.a))
```

I have add docs and tests.

### DIFF
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -275,3 +275,61 @@ function show_composition_order(io::IO, lens::ComposedOptic)
     print(io, ")")
 end
 
+
+"""
+    decompose(optic)
+
+decomposes the `ComposedOptic` into tuple of simple properties (Lenses) 
+while preserving the order
+
+```julia 
+julia> decompose(@optic _.a.b.c.d)
+((@optic _.d), (@optic _.c), (@optic _.b), (@optic _.a))
+
+julia> decompose((@optic _.c.d) ∘ (@optic _.a.b))
+((@optic _.d), (@optic _.c), (@optic _.b), (@optic _.a))
+"""
+decompose(optic) = (optic,)
+decompose(optic::ComposedOptic) = (decompose(optic.outer)..., decompose(optic.inner)...)
+
+"""
+    propname(::PropertyLens{name})
+
+returns the name of the propery lens 
+```julia
+julia> propname(@optic _.a)
+:a
+```
+
+Using `propname` with `decompose`, list all keys of can be listed as 
+```julia
+julia> map(propname, decompose((@optic _.c.d) ∘ (@optic _.a.b))
+(:d, :c, :b, :a)
+```
+"""
+propname(::PropertyLens{name}) where {name} = name
+
+
+"""
+    normalise(optic)
+
+transforms the optic such that the `inner` lens contains always a the `PropertyLens`
+
+```julia 
+julia> optic = (@optic _.a) ∘ (@optic _.b.c)
+(@optic _.a) ∘ (@optic _.c) ∘ (@optic _.b)
+
+julia> optic.inner
+(@optic _.c) ∘ (@optic _.b)
+
+julia> optic.outer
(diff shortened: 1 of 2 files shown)


jw3126:
Thanks a lot @pevnak ! What do you think @tkf ?

* I think `decompose` should go to `CompositionsBase`. What it does is
`decompose(sin ∘cos ∘ tan) === (sin, cos, tan)`.

* `normalize` does `normalize((sin ∘ cos) ∘(tan ∘ cot)) === sin ∘ cos ∘ tan ∘ cos`. It could also go to `CompositionsBase`. 



https://github.com/jw3126/Setfield.jl/issues/150

pevnak:
U know we have talked about it, I just did not know, if it fits there an where?

Should I move it?

jw3126:
Not sure if you are aware. But `ComposedOptic` is just an alias for `ComposedFunction`.
https://github.com/JuliaObjects/Accessors.jl/blob/master/src/optics.jl#L90

So if you define `decompose(f::ComposedFunction) = ...` then composed lenses will automatically work. 
So you can just put `decompose(...)` at the bottom of https://github.com/JuliaFunctional/CompositionsBase.jl/blob/master/src/CompositionsBase.jl

About `normalize`, let's wait if @tkf has an opinion on where it should live.

jw3126:
@pevnak lets keep normalization here and move `decompose` to CompositionsBase.jl

src/sugar.jl
```diff
@@ -275,3 +275,63 @@ function show_composition_order(io::IO, lens::ComposedOptic)
...
+```
+"""
+decompose(optic) = (optic,)
+decompose(optic::ComposedOptic) = (decompose(optic.outer)..., decompose(optic.inner)...)
+
+"""
+    propname(::PropertyLens{name})
+
+returns the name of the propery lens 
+```julia
```
jw3126, line 300:
```suggestion
```jldoctest```
Similar below.

File: src/sugar.jl [hunk 2]
```diff
@@ -275,3 +275,63 @@ function show_composition_order(io::IO, lens::ComposedOptic)
...
+julia> map(propname, decompose((@optic _.c.d) ∘ (@optic _.a.b))
+(:d, :c, :b, :a)
+```
+"""
+propname(::PropertyLens{name}) where {name} = name
+
+
+"""
+    normalise(optic)
```
jw3126, line 315: Lets spell it the American way `normalize` to be consistent with julia base.

src/sugar.jl
```diff
@@ -275,3 +275,63 @@ function show_composition_order(io::IO, lens::ComposedOptic)
     print(io, ")")
 end
 
+
+"""
+    decompose(optic)
```
jw3126, line 280: Should be moved to CompositionsBase.jl

jw3126 (changes requested):
Thanks looks good overall, just some minor remarks. Also, we should add a test, that normalize yields the compiler preferred composition order (I think this might be violated with your current implementation):
https://github.com/JuliaObjects/Accessors.jl/blob/master/test/perf.jl

jw3126:
https://github.com/JuliaFunctional/CompositionsBase.jl/pull/9

jw3126:
@pevnak no idea if this is still relevant for you, but I added decompose here https://github.com/JuliaFunctional/CompositionsBase.jl/pull/9

PR #22, opened 2021-04-04
setindex for ref

balenamiaa:


src/setindex.jl
```diff
@@ -20,3 +20,9 @@ Base.@propagate_inbounds function setindex(d0::AbstractDict, v, k)
     d[k] = v
     return d
 end
+
+Base.@propagate_inbounds function setindex(ref::Base.RefValue{Tx}, val::Ty) where {Tx, Ty}
+    T = promote_type(Tx, Ty)
+    new_ref = Base.RefValue{T}(convert(T, val))
+    return new_ref
+end
```
jw3126, line 28: Why the `Base.@propagate_inbounds`? Also I think I would prefer not to promote here. I think just 
```julia
setindex(::Base.RefValue, val) = Ref(val)
```
is fine. Promotion can quickly lead to eltype `Any`.

jw3126 (commented):
Thanks for the PR! Can you add some tests? See also my comment above.

src/setindex.jl
jw3126: For other containers we do promote in `setindex`. However these containers typically have multiple elements and so the new container must contain both old elements and the new one added. A `Ref` however has always only a single element, so there is less need to ensure the old element can still be contained.

balenamiaa: I think the function should be force inlined, so it doesn't allocate if the original Ref was stack allocated, though `Base.@propagate_inbounds` isn't required since there is no indexing. Also, I am sure by default it'll get inlined by the compiler so I don't know if the explicitness is required.

Regarding promotion, you're right, there's no need for promotion or conversion.

balenamiaa:
Oh, it seems escape-analysis fails in Julia 1.3 therefore the test for 0-allocation fails too. Should I remove the allocation test?

test/test_setindex.jl
```diff
@@ -32,6 +40,15 @@ end
     @test d == Dict(:a => 1, :b => 2)
     @test Accessors.setindex(d, 30, "c") ==ₜ Dict(:a=>1, :b=>2, "c"=>30)
     @test Accessors.setindex(d, 10.0, :a) ==ₜ Dict(:a=>10.0, :b=>2.0)
+    
+    ref = Ref((; a = 10, b = 0x0001, c = (; aa = 300)))
+    @test @set(ref[].a = 90)[] == Ref((; a = 90, b = 0x0001, c = (; aa = 300)))[]
```
jw3126:
```suggestion
    @test @set(ref[].a = 90)[] == (; a = 90, b = 0x0001, c = (; aa = 300))
```
and similar below.

jw3126: Also, can you add a test, that ref was not mutated?

jw3126:
> Oh, it seems escape-analysis fails in Julia 1.3 therefore the test for 0-allocation fails too. Should I remove the allocation test?

No, just warp it with something like `if VERSION >= v"1.5"`

test/test_setindex.jl
```diff
@@ -32,6 +40,19 @@ end
     @test d == Dict(:a => 1, :b => 2)
     @test Accessors.setindex(d, 30, "c") ==ₜ Dict(:a=>1, :b=>2, "c"=>30)
     @test Accessors.setindex(d, 10.0, :a) ==ₜ Dict(:a=>10.0, :b=>2.0)
+    
+    ref = Ref((; a = 10, b = 0x0001, c = (; aa = 300)))
+    @test @set(ref[].a = 90)[] ==ₜ (; a = 90, b = 0x0001, c = (; aa = 300))
+    @test @set(ref[].b = Bool(ref[].b))[] == (; a = 10, b = true, c = (; aa = 300))
+    @test @set(ref[].c.aa = 3)[] == (; a = 10, b = 0x0001, c = (; aa = 3))
+
```
jw3126, line 48:
```suggestion

ref = Ref(1)
@set ref[] = "no mutation"
@test ref[] === 1

ref = Ref(1)
@test typeof(@set ref[] = :other_type) == typeof(Ref(:other_type))

```

balenamiaa:
The CI checks should fail as there seems to be a type instability in `Accessors.modify`.
<pre>
julia> @code_warntype ref_alloc_test()
Variables
  #self#::Core.Const(ref_alloc_test)
  ref2::<b>Base.RefValue{_A} where _A</b>
  ##_#288::Base.RefValue{_A} where _A
  optic::ComposedFunction{Accessors.IndexLens{Tuple{Int64}}, Accessors.IndexLens{Tuple{}}}
  ref::Base.RefValue{Tuple{UInt64, Int64}}

Body::<b>Any</b>
1 ─ %1  = Main.UInt(10)::Core.Const(0x000000000000000a)
│   %2  = Core.tuple(%1, 100)::Core.Const((0x000000000000000a, 100))
│         (ref = Main.Ref(%2))
│   %4  = ()::Core.Const(())
│   %5  = (Accessors.IndexLens)(%4)::Core.Const((@optic _[]))
│   %6  = Core.tuple(1)::Core.Const((1,))
│   %7  = (Accessors.IndexLens)(%6)::Core.Const((@optic _[1]))
│   %8  = (Accessors.opticcompose)(%5, %7)::Core.Const((@optic _[1]) ∘ (@optic _[]))
│         (optic = (identity)(%8))
│   %10 = (Accessors._UpdateOp)(Accessors.:-, 20)::Core.Const(Accessors._UpdateOp{typeof(-), Int64}(-, 20))
│   %11 = ref::Base.RefValue{Tuple{UInt64, Int64}}
│   %12 = (Accessors.modify)(%10, %11, optic::Core.Const((@optic _[1]) ∘ (@optic _[])))::<b>Base.RefValue{_A} where _A</b>
│         (##_#288 = %12)
│         (ref2 = %12)
│   %15 = Base.getindex(ref2)::<b>Any</b>
└──       return %15
</pre>

balenamiaa:
```
Body::Base.RefValue{_A} where _A
│    @ ~/Accessors.jl/src/optics.jl:140 within `_set'
│   ┌ @ ~/Accessors.jl/src/optics.jl:336 within `set'
│   │┌ @ ~/Accessors.jl/src/setindex.jl:2 within `setindex' @ tuple.jl:50
│   ││ %13 = Base.BoundsError(%1, %4)::Any
│   ││ @ ~/Accessors.jl/src/setindex.jl:2 within `setindex' @ tuple.jl:52
│   ││┌ @ tuple.jl:57 within `_setindex'
│   │││ %19 = Base.ifelse(%18, val, %16)::Union{Int64, String}
│   │││ @ tuple.jl:57 within `_setindex' @ tuple.jl:57
│   │││ %22 = Base.ifelse(%21, val, %17)::Union{Int64, String}
│   │││ @ tuple.jl:57 within `_setindex'
│   │││ %23 = Core.tuple(%19, %22)::Tuple{Union{Int64, String}, Union{Int64, String}}
│   └└└
│    @ ~/Accessors.jl/src/optics.jl:141 within `_set'
│   ┌ @ ~/Accessors.jl/src/optics.jl:336 within `set'
│   │┌ @ ~/Accessors.jl/src/setindex.jl:5 within `setindex'
│   ││ %27 = Accessors.Ref(%23)::Base.RefValue{_A} where _A
    └└
```
Basically tuple setindex defined in base is like so: `setindex(::Tuple{Int64, Int64},::String,::Int64)::Tuple{Union{Int64, String}, Union{Int64, String}}`. I am not exactly sure how to solve this but it doesn't happen for named tuples, e.g:
```
obj = Ref((; a = 10, b = 20))
optic = @optic _[].a
val = "fdfdf"
@code_warntype Accessors.set(obj, optic, val)

optic_style = Accessors.OpticStyle(optic)
@code_warntype Accessors._set(obj, optic, val, optic_style)

function f()
    inner_obj = optic.inner(obj)
    inner_val = set(inner_obj, optic.outer, val)
    inner = optic.inner
    @code_warntype Accessors.set(obj, inner, inner_val)
end
``` results in 
```
Variables
  #self#::Core.Const(Accessors.set)
  obj::Base.RefValue{NamedTuple{(:a, :b), Tuple{Int64, Int64}}}
  optic::Core.Const((@optic _.a) ∘ (@optic _[]))
  val::String

Body::Base.RefValue{NamedTuple{(:a, :b), Tuple{String, Int64}}}
1 ─      nothing
│   %2 = Accessors.OpticStyle(optic)::Core.Const(Accessors.SetBased())
│   %3 = Accessors._set(obj, optic, val, %2)::Base.RefValue{NamedTuple{(:a, :b), Tuple{String, Int64}}}
└──      return %3
Variables
  #self#::Core.Const(Accessors._set)
  obj::Base.RefValue{NamedTuple{(:a, :b), Tuple{Int64, Int64}}}
  optic::Core.Const((@optic _.a) ∘ (@optic _[]))
  val::String
  #unused#::Core.Const(Accessors.SetBased())
  inner_val::NamedTuple{(:a, :b), Tuple{String, Int64}}
  inner_obj::NamedTuple{(:a, :b), Tuple{Int64, Int64}}

Body::Base.RefValue{NamedTuple{(:a, :b), Tuple{String, Int64}}}
1 ─      nothing
│   %2 = Base.getproperty(optic, :inner)::Core.Const((@optic _[]))
│        (inner_obj = (%2)(obj))
│   %4 = inner_obj::NamedTuple{(:a, :b), Tuple{Int64, Int64}}
│   %5 = Base.getproperty(optic, :outer)::Core.Const((@optic _.a))
│        (inner_val = Accessors.set(%4, %5, val))
│   %7 = Base.getproperty(optic, :inner)::Core.Const((@optic _[]))
│   %8 = Accessors.set(obj, %7, inner_val)::Base.RefValue{NamedTuple{(:a, :b), Tuple{String, Int64}}}
└──      return %8
```. No type instabilities for NamedTuple. Is this a bug?

balenamiaa:
Perhaps this should be a separate issue here as this isn't related to Ref setindex.

balenamiaa:
There are no type instabilities, they get constant propagated away it seems. Still, using modify allocates for a tuple(and not a named tuple), meanwhile set does not allocate(even if the type changes). It seems that one of the modify methods were not inlined, will send a PR soon.

jw3126:
Somehow it seems you changed every file in the project. Can you restore unrelated files please?

jw3126 (approved):
LGTM! Thanks @balenamiaa ! Is this PR ready to merge from your side?

jw3126:
@balenamiaa if you bump the minor version, I can immediately tag a release after merge.

Issue #9, opened 2020-10-25
Differences to Setfield.jl?


rafaqz:
This package is great to see. Just wondering if you have a list of differences and changes from Setfield.jl - current or intended?

jw3126:
There is an incomplete list of changes [here](https://github.com/jw3126/Setfield.jl/pull/147). Some goals were to make lenses more light weight, in particular duck typed and avoid overloading `Base.get`.

Can you given an example what kind of `Flatten` like thing you would like to do? I think for modification, this package should be quite powerful e.g. replace all orange leaves by red ones in a tree. For actual flattening e.g. collect all orange leaves into a vector, this is currently not supported. And I am not certain what is the best design here. Would be interested in your input.

rafaqz:
My main practical use case now is https://github.com/rafaqz/ModelParameters.jl

As mentioned the main thing missing is context about the leaves on the tree that are replaced. What is their parent object, what is the field name, how to call a function of those two things (with fieldname in `Val{:fieldname}` for dispatch).

rafaqz:
This is most of Flatten.jl rI rewrote a while ago for Setfield.jl lenses, with `Context` (just the `get` half):

```julia
using Setfield

using Setfield: ComposedLens

struct SkipNone end

abstract type Query{S,R} <: Lens end

struct Select{S,R} <: Query{S,R} end
Select{S}() where {S} = Select{S,SkipNone}()

struct Context{X,S,R} <: Query{S,R} end
Context{X,S}() where {X,S} = Context{X,S,SkipNone}()

struct FieldName end
struct ParentObj end
struct ParentType end

@inline query(lens::Select, obj, ::Val{FN}, val) where FN = (val,)
@inline query(lens::ComposedLens, obj, fn::Val{FN}, val) where FN =
    (get(query(lens.outer, obj, fn, val)[1], lens.inner),)
@inline query(lens::Context{FieldName}, obj, ::Val{FN}, val) where FN = (FN,)
@inline query(lens::Context{ParentObj}, obj, ::Val{FN}, val) where FN = (obj,)
@inline query(lens::Context{ParentType}, obj, ::Val{FN}, val) where FN = (typeof(obj),)

@inline Setfield.get(obj, lens::Union{ComposedLens{<:Query},Query}) = _get(obj, lens)
@inline Setfield.set(obj, lens::Union{ComposedLens{<:Query},Query}, x) = _set(obj, lens, x)

[...]

PR #23, opened 2021-06-06
first pass get/set queries

rafaqz:
This PR adds get/set queries as discussed in https://github.com/JuliaObjects/Accessors.jl/issues/9, returning a `Tuple` or setting values from a `Tuple` or anything that can be indexed.

Currently you can select objects by a `select` function, and ignore other objects with an `ignore` function. `modify` seems to compile away even if these functions are not type stable for `set`.

Composing with other lenses doesn't work yet, it's a little more complicated than with Setfield.jl as Accessors is using functors instead of 2-arg `get` where you could dispatch on the composed object in `get`. Any ideas here?

For macros I was thinking something like this could be nice:

```julia
@getall x isa Int
@setall x isa AbstractArray = arrays
```

TODO:
- [ ] context queries, where the returned value can be some function of obj/fieldname/value
- [ ] tree context queries, where you can e.g. build a tree structure of e.g. fieldnames that lead to the queries objects
- [ ] include `Properties/Elements` distinction, and maybe add `Fields`? this is using fieldnames in a `@generated` function


src/optics.jl
```diff
@@ -272,7 +292,72 @@ $EXPERIMENTAL
...
 OpticStyle(::Type{<:Properties}) = ModifyBased()
-modify(f, o, ::Properties) = mapproperties(f, o)
+function modify(f, o, ::Properties, handler=Construct())
+    mapproperties(f, o, handler)
+end
+
+"""
+    mapfields(f, obj)
+
+Construct a copy of `obj`, with each property replaced by
```
jw3126:
```suggestion
Construct a copy of `obj`, with each field replaced by
```

src/optics.jl
```diff
@@ -185,7 +205,7 @@ $EXPERIMENTAL
 struct Elements end
 OpticStyle(::Type{<:Elements}) = ModifyBased()
 
-function modify(f, obj, ::Elements)
+function modify(f, obj, ::Elements, object=List())
```
jw3126: There are other ideas what n argument modify could mean. I think defining it is not essential for this PR? So let's try to avoid it.

jw3126: Alternatively if you feel strongly about using `modify` we can also discuss that.

jw3126 (commented):
Wow, thanks a lot. I find the code super elegant, especially how you could "contain" the use of `@generated`! Does it infer well? Does composition work?

src/optics.jl
rafaqz: The reason for adding this argument is so we can use the same code to both `Construct` new objects in and return tuples of fields, with `Tuple` (for all) and `Splat` (for some filtered subset). This is how Flatten.jl works, it's just impossible to read that that is whats happening - your structure is so much cleaner here.

We can also use a different method entirely to do this, this was just me trying to find the cleanest way of using everything already written here and build things so the other optics can use what Ive added.

rafaqz: This method isn't working currently for that, it's just a dummy argument so it doesn't break.

rafaqz:
Thanks! It was awful before I rewrote it with the structure here.

I'm not sure how to do composition properly yet, it needs more thinking. That's the part I understand the least about Accessors.jl so far... but getting there.

rafaqz:
The iterator is broken too, I knew it couldn't be that easy lol...

src/optics.jl
jw3126: Ok sure, we can in the end decide whether to rename this. Competing ideas were to use extra args of modify for passing context around. Or have `modify(f, obj, lens)` behave to `map(f, itr)` as `modify(f, objs..., lens)` behaves to `map(f, itrs...)`.

rafaqz: Better to keep the syntax open.

I've made them separate methods for now, we can work towards reducing the differences. Probably it should only affect the inner `_modify` not `modify` anyway.

rafaqz:
This is composing now. It hast to compose internally, as you want to compose with whatever fields are selected by Query, not with its return value.

jw3126:
> This is composing now. It hast to compose internally, as you want to compose with whatever fields are selected by Query, not with its return value.

Can you give a few examples of what Query should do and how you think it should compose?
I toyed a bit with this branch and found that things often do not behave as I anticipated.
Now I realize that this is very much WIP so I am not sure, whether I don't understand the intended semantics of Query or whether these are just bugs / not really implemented.

For instance
```julia
julia> using Accessors

julia> obj = (a=missing, b=1, c=(d=missing, e=(f=missing, g=2)))
(a = missing, b = 1, c = (d = missing, e = (f = missing, g = 2)))

julia> Query(ismissing)(obj)
(missing, 1, missing, missing, 2)
```
I would have expected `(missing,missing,missing)` instead.

Also are querries supposed to be lenses? That is should `q::Query` satisfy the lens laws:
```julia
@assert q(set(obj, q, val)) ≅ val
        # You get what you set.
@assert set(obj, q, q(obj)) ≅ obj
        # Setting what was already there changes nothing.
@assert set(set(obj, q, val1), q, val2) ≅ set(obj, q, val2)
        # The last set wins.
```

rafaqz:
Yeah that's just a bug, the answer should be what you expect. I need to write a few more tests.

It should follow the lens laws... notice modify just uses `map` over the results of get to feed into set:
https://github.com/JuliaObjects/Accessors.jl/pull/23/files#diff-705e67cd19daff66a5500d5e0cab028615f413efbd83f9420030c9cd443e62ffR554

rafaqz:
Ok this version should be working for lens laws, but with some problems with set for `Properties` and bad set performance.

jw3126:
I just realized, that querries are not lenses in general.
```julia
@assert q(set(obj, q, val)) ≅ val
        # You get what you set.
@assert set(obj, q, q(obj)) ≅ obj
        # Setting what was already there changes nothing.
@assert set(set(obj, q, val1), q, val2) ≅ set(obj, q, val2)
        # The last set wins.
```
I believe the first and second lens law should always hold, but the third may not:
```julia
julia> using Accessors

julia> obj = (a=missing, b=1, c=(d=missing, e=(f=missing, g=2)))
(a = missing, b = 1, c = (d = missing, e = (f = missing, g = 2)))

julia> q = Query(ismissing);

julia> obj2 = set(obj, q, ["first", "wins", "here"])

julia> set(obj2, q, ["second", "should", "win"])
```
Not a huge deal, but we should be careful about using `lens` as a variable name.

src/optics.jl
```diff
@@ -229,50 +278,106 @@ julia> using Accessors
...
-julia> modify(x -> 2x, obj, Properties())
+julia> modify(x -> 2x, obj, Fields())
 (a = 2, b = 4, c = 6)
 ```
-Based on [`mapproperties`](@ref).
+Based on [`mapfields`](@ref).
 
 $EXPERIMENTAL
 """
-struct Properties end
```
jw3126, line 273: Since we have `ConstructionBase.getproperties` and `ConstructionBase.setproperties` I wonder if we could base this on
```julia
set(obj, ::Properties, val) = setproperties(obj, val)
(::Properties)(obj) = getproperties(obj)
```

rafaqz: Yeah that's probably better.

rafaqz:
Yeah... the query would have to be updated for that to hold, e.g:

```julia
julia> set(obj2, Query(x -> x isa String), ["second", "should", "win"])
(a = "second", b = 1, c = (d = "should", e = (f = "win", g = 2)))
```

So it breaks 3. when the the values passed to `set` would give a different query result to the values in the object that they replace. We just need to give that a name I guess.

Also let me know what you think of the macros (theyre in the tests now too). They seem nice for simple use cases, just using the Query object is probably easier beyond something like this:

```julia
@getall missings_obj isa Number
@setall missings_obj isa Number = (5, 6)
```

rafaqz:
How do you feel about a dependency on [Stattic.jl](https://github.com/SciML/Static.jl/blob/master/src/int.jl)?

Using `StaticInt` as an iterator solves the type stability with `set`.

jw3126:
> How do you feel about a dependency on [Static.jl](https://github.com/SciML/Static.jl/blob/master/src/int.jl)?
> 
> Using `StaticInt` as an iterator solves the type stability with `set`.

`Static.jl` looks like a small dependency, so I am fine with it.

rafaqz:
Turns out Static.jl doesn't actually fix it, I'm just getting Revise.jl artifacts from this Julia bug:
https://github.com/JuliaLang/julia/issues/35800

We'll have to settle for 20-100ns or so timings until that gets fixed. I can get everything under 2ms using Revise but it wont stick on restart, which has been driving me crazy.

jw3126:
I think I have an alternative approach how this can be implemented. Let me first sketch a slow implementation,
analyze why it is slow and sketch how to fix it.
```julia
struct Query{S,D,O}
    select::S
    descend_condition::D
    optic::O
end

function modify(f, obj, q)
    modify(obj, q.optic) do o
        if q.select_condition(o)
            f(o)
        elseif q.descent_condition(o)
            modify(f, o, q)
        else
            o
        end
    end
end

pop(x) = first(x), Base.tail(x)
push(x, val) = (x..., val)

function setall(obj, q, vals)
    modify(obj, q) do o
        val, vals=pop(vals)
        val
    end
end

function getall(obj, q)
    ret = ()
    modify(obj,q) do o
        push(ret, o)
        o
    end
    return ret
end
```
Now this is slow, because we change captured variables in a closure. How to fix this?
By passing state around explicitly. So we have to replace `modify` as a primitive by a state aware `modify_stateful`.

```julia
function getall(obj, q)
    initial_state = ()
    _, final_state = modify_stateful((obj, initial_state), q) do o, state
        new_state = push(state, o)
        o, new_state
    end
    return final_state
end

"""

    new_obj, new_state = modify_stateful(f, (obj,state), optic)

Here `f` has signature `f(::Value, ::State) -> Tuple{NewValue, NewState}`.
"""
function modify_stateful end

function modify_stateful(f, (obj, state), optic::Properties)
    # @generated function that produces the following code:
    obj0 = getproperties(obj)
    obj_state0 = (obj0, state)
    obj_state1 = modify_stateful(f, obj_state0, optic)
    obj_state2 = modify_stateful(f, obj_state1, optic)
    obj_state3 = modify_stateful(f, obj_state2, optic)
    obj_state4 = modify_stateful(f, obj_state3, optic)
    obj_state5 = modify_stateful(f, obj_state4, optic)
    new_nt, new_state = obj_state5
    new_obj = setproperties(obj, new_nt)
    return (new_obj, new_state)
end

function modify_stateful(f, (obj, state), q::Query)
    modify_stateful((obj,state), q.optic) do o,s
        if q.select_condition(o)
            f(o,s)
        elseif q.descent_condition(o)
            modify_stateful(f, (o,s), q)
        else
            o, s
        end
    end
end

function modify_stateful(f, (obj, state), ::ComposedOptic)
    ...
end

# all currently existing modify methods in Accessors.jl can be rewritten into `modify_stateful` in a straight forward fashion.
# Like with modify we want to use trait based dispatch `_modify_stateful(f, obj_state, optic, ::OpticStyle)`.
# `modify` may in the end be reimplemented in terms of `modify_stateful` if the compiler likes it.
```
What do you think?

rafaqz:
Wow that looks like a flexible generalisation, lets use it. Thanks for putting the time in to rethink this.

jw3126:
> Also let me know what you think of the macros (theyre in the tests now too). They seem nice for simple use cases, just using the Query object is probably easier beyond something like this:

I wonder if we can capture generator syntax here to make the macros a bit more flexible.
```julia
@getall (x for x in obj if select(x))
@getall Float64[x for x in obj if select(x)] # collects into a vector of Float64 
@modify CuArray(x) for x in obj if x isa AbstractArray # move everything to GPU
@getall (x for x in obj |> Elements()) # customize the optic to flatten a nested tuple
```

rafaqz:
This is nice syntax:
```julia
@getall Float64[x for x in obj if select(x)] # collects into a vector of Float64 
```

You could also use lenses:

```julia
@getall [x.a for x in obj if select(x)]
@getall (x[2] for x in obj if x isa NamedTuple)
```

jw3126:
Not for this PR, but for a later stage I think we can connect this with Transducers. The idea is to turn `(obj, optic)` into a [ReducibleCollection](https://juliafolds.github.io/Transducers.jl/stable/explanation/glossary/#Reducible-collection).
Roughly speaking by:
```julia
function Transducers.foldl(rf, ReducibleCollection(obj, optic), init)
    # TODO this is the wrong method to overload, but you get the idea
    _, ret = modify_stateful((obj, init), optic) do o, state
        new_state = rf(state, o)
        o, new_state
    end
    return ret
end
```

We would have:
```julia
getall(obj, q) = foldl(push!!, ReducibleCollection(obj, q))
```
Benefits would be that we could directly apply reduction functions to queries and Transducers could take care of the output container (e.g. Tuple, Vector, lazy iterator...).

rafaqz:
Transducers interop sounds good for a future PR. I probably have to leave that to you as I haven't had time to get my head around transducers yet.

jw3126:
> You could also use lenses:
> 
> ```julia
> @getall [x.a for x in obj if x isa NamedTuple]
> @setall (x[2] for x in obj if x isa AbstractArray) = (1, 2 ,3)
> ```

Ah nice, this is a way to deal with simple "contextful" queries.

rafaqz:
The context that is harder to get is the parent object and field name (probably as a `StaticSymbol/StaticInt`). Fieldname is really useful for e.g. making `name => value` pairs e.g widget labels or checking you are querying the right fields. We can work it into `state` somehow.

This was fairly straightforward but not finished:
```julia
julia> missings_obj = (a=missing, b=1, c=(d=missing, e=(f=missing, g=2)))
(a = missing, b = 1, c = (d = missing, e = (f = missing, g = 2)))

julia> @getall (x.e for x in missings_obj if x isa NamedTuple)
((f = missing, g = 2),)
```

jw3126:
> The context that is harder to get is the parent object and field name (probably as a `StaticSymbol/StaticInt`). Fieldname is really useful for making `name => value` pairs e.g widget labels or checking you are querying the right fields. We can work it into `state` somehow.

I totally agree fieldnames are so useful. More general context is tough and IMO out of scope for this PR. But as you say the way to go piggy bag context into state. Probably via some `PushCtx()` mechanism like what I tried to sketch [here](https://github.com/JuliaObjects/Accessors.jl/issues/9#issuecomment-722197946).

rafaqz:
I've implemented `modify_stateful` and built `getall`, `setall` and `context` methods on it generally following the ideas from [this comment](https://github.com/JuliaObjects/Accessors.jl/pull/23#issuecomment-857360374). The structure seems pretty clean, although the generated function is still a bit of a mess. State wrappers like `ConstextState`, `GetAllState` and `SetAllState` allow dispatching behaviours for each type and also wrapping state objects.

Unfortunately it's harder to optimise now! I've got some lingering type instabilities. Maybe you can see whats causing them.

src/optics.jl
```diff
@@ -318,6 +337,189 @@ function _modify(f, obj, r::Recursive, ::ModifyBased)
...
+        modify_stateful((obj, state), inner(q.optic)) do o, s
+            if (q::Q).select_condition(o)
+                (f::F)(o, s)
+            elseif (q::Q).descent_condition(o)
+                ds = descent_state(s)
+                o, ns = modify_stateful(f::F, (o, ds), q::Q)
+                o, merge_state(ds, ns)
+            else
+                o, s
+            end
```
jw3126, line 491: Thanks a lot for pushing this further. I also played around with implementing this, see this branch:
https://github.com/JuliaObjects/Accessors.jl/tree/modify_stateful  
Like with your implementation I encountered inference issues. I do not fully understand them, but it seems the compiler dislikes this switch. To explore what's going on I replaced `elseif (q::Q).descent_condition(o)` into  `elseif true` but the thing would still not infer. OTOH manually deleting the `elseif` branch did help the compiler to infer.

Note I did this experimentation with my branch, which is somewhat different then yours, but I guess has the same underlying issues. Worth checking if things are better with the new ConstructionBase version and nightly julia.

If not I think it would be good to produce an MWE and report it upstream. I don't know if it is considered a bug, but maybe we'll get an explanation of why it is this way and how to work around it.

A quick and dirty workaround would be to introduce a more narrow query where you can only select and descent based on type. Then the `if elseif ...` could be replaced by three methods.

rafaqz: Ok interesting we both hit the same problem. 

I think the differences are mostly just my code to handle context and optional construction in setall. But those aren't whats causing the instability... 

It seems like the problem is calling `modify_stateful` recursively? It still seems broken even without the `elseif`, and the `elseif` was fine in the previous version. I don't understand exactly what is different between these implementations and my previous one or Flatten.jl, but it could be the additional complexity of the return objects? Or there is a little more happening at runtime in `modify_stateful` than I had in `mapobject` or Flatten.jl, and the compiler is giving up on something?

I also saw no difference between ConstructionBase versions.

rafaqz: Seems its the layers of anonymous functions. Flatten.jl has the [shared code](https://github.com/rafaqz/Flatten.jl/blob/master/src/Flatten.jl#L29-L33) inside the generated functions, and simple [separate generated functions](https://github.com/rafaqz/Flatten.jl/blob/master/src/Flatten.jl#L124-L131) for getall/setall (flatten/reconstruct).

I much prefer the clarity of the current solution here. But it seems like if we want this to compile away we cant use anonymous functions inside the recursion.

jw3126: It does not seem to be related to anonymous functions per see. I played with eliminating all anonymous functions.
I replaced code like
```julia
modify_stateful(...) do os
    do_stuff(os, captured, variables)
end
```
with code like
```julia
struct DoStuff{C,V}
    captured::C
    variables::V
end
(f::DoStruff)(os) = do_stuff(os, f.captured, f.variables)
modify_stateful(DoStuff(captured, variables), ...)
```
But it helped in very small examples, small examples still fail. Also, `@inline` everything did not help.

rafaqz: Damn, that's exhaustive.

It feels like the compiler is using some heuristic to give up on solving the returned state of the recursion instead of just trying. A single non-recursive `modify_stateful` compiles away completely, but it wont if called recursively.

rafaqz: Ok, think i found the issue. If you define the select and descent conditions like this it's type stable:

```julia
select_condition(x::NamedTuple) = true
select_condition(x) = false

descent_condition(x::Tuple) = true
descent_condition(x) = false
```

So the compiler optimises these methods differently to methods that are passed in in any way. Even with the function in the  type instead of fields, or just passing a Type to a 2 arg version of `descend_condition` still doesn't work. It's just giving up and returning `Any`. If you actually include the types to descend on in the code, it's type-stable. 

I had assumed previously that using a type from a type parameters was practically the same as writing the type in the code manually, but it seems like it isn't.

jw3126: Awesome that you found the issue! This looks like a bug to me, can you maybe produce a MWE and report it upstream?

rafaqz: Yeah I'm cutting it down to size, but its probably 1.6 inference problems like JuliaLang/julia#35800.

It's all type stable on 1.4 - that's why this feels weird for performance intuition. These things used to just work.

rafaqz: Eh only some things are more stable on 1.4. #35800 just has me not trusting the compiler anymore.

rafaqz:
@jw3126 sorry I haven't had time for this in a while, just revisiting now. 

How do you feel about these options:

1. finish this with the current type instability
2. reduce the scope to only matching types (so all at compile time) and separate out more generated functions to get type stability

jw3126:
> @jw3126 sorry I haven't had time for this in a while, just revisiting now.
> 
> How do you feel about these options:
> 
>     1. finish this with the current type instability
> 
>     2. reduce the scope to only matching types (so all at compile time) and separate out more generated functions to get type stability

How urgent is this for you? One option would be to just wait until either inference can handle this out of the box
or compiler plugins are stable and powerful enough that we can help inference.
If you don't want to wait that long, I am fine with 1. About 2. my concern is that implementation will be complicated and composability will be a problem.

rafaqz:
Ok maybe we can do 1 and fix the performance things when we can.

The (low key) urgency is that I'm keen to move on from using Flatten.jl in a few packages. The lack of checks on state change put a ceiling on its use when there are fields in an object that don't have `constructorof` defined, and if I'm adding functionality like that I'd rather it was here than in Flatten.jl.

jw3126:
> Ok maybe we can do 1 and fix the performance things when we can.
> 
Sounds good!

rafaqz:
Ok this is pretty much good to go. The last question is what to do with the `context` method and how to write `@modify` in relation to the generator syntax of the `@setall` and `@getall` macros.

src/optics.jl
```diff
@@ -318,6 +323,185 @@ function _modify(f, obj, r::Recursive, ::ModifyBased)
...
+            new_state = push(s, outer(q.optic, o, s))
+            o, new_state
+        end
+    end
+    final_state.vals
+end
+
+function set(obj, q::Q, vals) where Q<:Query
+    initial_state = SetAllState(Unchanged(), vals, 1)
+    final_obj, _ = let obj=obj, q=q, initial_state=initial_state
```
jw3126, line 439: These variables are not captured, but they are function arguments directly. So no need for let block?

src/optics.jl
```diff
@@ -318,6 +323,185 @@ function _modify(f, obj, r::Recursive, ::ModifyBased)
...
+@inline pop(x) = first(x), Base.tail(x)
+@inline push(x, val) = (x..., val)
+@inline push(x::GetAllState, val) = GetAllState(push(x.vals, val))
+@inline push(x::ContextState, val) = ContextState(push(x.vals, val), nothing, nothing)
+
+(q::Query)(obj) = _getall(obj, q)
+
+function _getall(obj, q::Q) where Q<:Query
+    initial_state = GetAllState(())
+    _, final_state = let q=q
```
jw3126, line 428: q is actually not captured in the closure.

src/optics.jl
```diff
@@ -318,6 +323,185 @@ function _modify(f, obj, r::Recursive, ::ModifyBased)
...
+        else
+            :state
+        end
+        ms = :(($v, state) = f(getfield(props, $(QuoteNode(fn))), $st))
+        push!(modifications, ms)
+        push!(vals.args, v)
+    end
+    patch = O <: Tuple ? vals : :(NamedTuple{$fns}($vals))
+    start = :(props = getproperties(obj))
+    rest = MacroTools.@q begin
```
jw3126, line 359: just for my education, is there an advantage of MacroTools.@q over quote here?

src/optics.jl
```diff
@@ -134,6 +135,10 @@ else
     using Base: Returns
 end
 
+struct Changed end
```
jw3126, line 138: How important is it, that we have `Changed/Unchanged`? It is a pure optimization and does not influence semantics, so we could drop it later at will if we wanted, right?

src/sugar.jl
```diff
@@ -84,13 +84,106 @@ end
...
+    elseif @capture(ex, (lens_ for var_ in obj_))
+        select = _ -> false
+        optic = _optics(lens)
+        :(Query($select, $descend, $optic)($(esc(obj))))
+    elseif @capture(ex, [lens_ for var_ in obj_])
+        select = _ -> false
+        optic = _optics(lens)
+        :([Query($select, $descend, $optic)($(esc(obj)))...])
+    else 
+        error("@getall must be passed a generator")
```
jw3126, line 137:
```suggestion
        error("@getall must be passed a generator or array comprehension")
```

jw3126 (commented):
Thanks a lot, @rafaqz for working on this!

src/optics.jl
jw3126: Ok that makes a ton of sense.

src/optics.jl
rafaqz: Maybe it has some effect on variable scope as well? I can't actually remember the reason now.

src/optics.jl
jw3126: ah right, q is captured.

bgroenks96:
Any updates on this?

jw3126:
I fear with the current compiler these type instabilities cannot be fixed. My feeling is the only way would be compiler plugins, but there is no mature way to do that. The other option is to merge a type unstable variant. Would that be useful to people?

bgroenks96:
Is it type unstable in all cases or just in specific cases?

jw3126:
IIRC queries of moderate complexity are not type stable, trivial queries work?

bgroenks96:
Do you mean the query complexity or the complexity of the type being queried?

If the former, having type stable `@getall x isa T` would already be enormously helpful.

jw3126:
The problem is recursion complexity. See also: https://github.com/JuliaLang/julia/issues/43296#issuecomment-991104427

rafaqz:
Oh wow so its actually getting worse, not better

jw3126:
> Oh wow so its actually getting worse, not better

Yeah, it got worse in 1.7.

rafaqz:
Ok, yes It seems like Flatten.jl also got worse with 1.7 - its a huge chunk of the TTFX for DynamicGrids.jl now.

What surprised me recently is that type instability in these methods seems to have a large effect on compile time - if you look at `@snoopi_deep` profiles the compiler chases the result type through the whole tree of recursive methods, but starting with abstract types. 

So it precompiles a whole tree of abstract methods that will never be used, which of course fails to produce a known output type but costs like half a second to conclude that. You cant precompile this away either.

bgroenks96:
So I guess I will continue sticking with 1.6 then...

bgroenks96:
> Ok, yes It seems like Flatten.jl also got worse with 1.7 - its a huge chunk of the TTFX for DynamicGrids.jl now.

Can you tell if it is `flatten` or `reconstruct` which is contributing the most?

rafaqz:
Pretty sure its `reconstruct`

bgroenks96:
Oof. Ok, then it might be the type tree actually causing it then.

rafaqz:
@bgroenks96 you might be right about generalising that package, but it is pretty annoying to do that work for the compiler.

Tokazama:
I can't be sure without explicitly testing this, but I think inference is failing because you are trying to operate on a complex type instead of the instance. I've encountered similar issues recently and you can get around it if you have a clear path through each struct's fields and are willing to do some work upfront with the instance instead of the type.

rafaqz:
Hmm maybe, but I'm not sure exactly what you mean. Do you have any specific lines/areas that could be changed?

It looked to me like the compiler gives up on chasing function return types through nested recursive functions. Which also happens on Flatten.jl now.

rafaqz:
Nice, looks similar to Flatten.jl.

Honestly I doubt this will ever be possible without the generated functions. 

So Im not sure what we will be waiting on to put it here...

jw3126:
> I noticed there's also a question of semantics: how should recursion work exactly. 

That is a very good point. We should support both. Naming is tricky though. Somewhat similar concepts are breadth first vs depth first search or `prewalk` and `postwalk` in MacroTools.

jw3126:
Happy new year! We could add the `modify/getall` part if you think it works well enough to Accessors.

PR #24, opened 2021-06-11
Properties

jw3126:
@rafaqz I hope I don't step on your toes with #23?


rafaqz:
Not at all! Feel free to implement any of your other ideas we discussed too, I think we are on the same page with all of this.

I'll be busy with other things for a while anyway so that PR wont be finished for a while. I'll just rebase this in.

PR #41, opened 2021-12-27
support multiple indices into NamedTuples

aplavin:
following new (julia 1.7) indexing behavior of `(a=1, b=2, c=3)[(:a, :b)] == (a=1, b=2)`

### DIFF
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -214,6 +214,13 @@ end
     l = @optic _[1:3]
     @test l isa Accessors.IndexLens
     @test l([4,5,6,7]) == [4,5,6]
+
+    nt = (a=1, b=2, c=3)
+    l = @optic _[(:a, :c)]
+    @test l isa Accessors.IndexLens
+    VERSION >= v"1.7" && @test l(nt) === (a=1, c=3)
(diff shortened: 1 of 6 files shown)


test/test_core.jl
```diff
@@ -214,6 +214,13 @@ end
     l = @optic _[1:3]
     @test l isa Accessors.IndexLens
     @test l([4,5,6,7]) == [4,5,6]
+
+    nt = (a=1, b=2, c=3)
+    l = @optic _[(:a, :c)]
+    @test l isa Accessors.IndexLens
+    VERSION >= v"1.7" && @test l(nt) == (a=1, c=3)
```
jw3126, line 221:
```suggestion
    VERSION >= v"1.7" && @test l(nt) === (a=1, c=3)
```
I think in places where `===` holds, we should test with `===`.

jw3126 (approved):
LGTM, thanks! Good work as usual. Feel free to merge and tag a release @aplavin

aplavin:
@jw3126 thanks! I don't think I can merge myself as I was unable to join your invitation: github requires 2fa there, and I don't really want that.

PR #43, opened 2022-01-16
remove assignments from non-overwriting macros

aplavin:
added tests that failed without this change


jw3126 (approved):
Thanks @aplavin! I was not even aware, that `[@set(...)]` did not work before!

PR #45, opened 2022-02-05
support two-argument functions in optics

aplavin:
For now, contains implementation and a single specific `delete` method as an example.
If others agree this extension is a good idea, will add more methods and tests.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Accessors"
 uuid = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 authors = ["Takafumi Arakaki <aka.tkf@gmail.com>", "Jan Weidner <jw3126@gmail.com> and contributors"]
-version = "0.1.8"
+version = "0.1.9"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
--- a/src/functionlenses.jl
+++ b/src/functionlenses.jl
@@ -7,6 +7,8 @@ delete(obj, ::typeof(first)) = delete(obj, IndexLens((firstindex(obj),)))
 insert(obj, ::typeof(last), val) = insert(obj, IndexLens((lastindex(obj) + 1,)), val)
 insert(obj, ::typeof(first), val) = insert(obj, IndexLens((firstindex(obj),)), val)
 
+delete(obj, o::Base.Fix2{typeof(first)}) = obj[(firstindex(obj) + o.x):end]
+
 set(obj, ::typeof(identity), val) = val
 set(obj, ::typeof(inv), new_inv) = inv(new_inv)
 
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -202,6 +202,24 @@ function parse_obj_optics(ex)
     elseif @capture(ex, f_(front_))
         obj, frontoptic = parse_obj_optics(front)
         optic = esc(f) # function optic
+    elseif @capture(ex, f_(args__))
+        args_contain_under = map(args) do arg
+            foldtree((yes, x) -> yes || x === :_, false, arg)
+        end
+        if !any(args_contain_under)
+            # as if f(args...) didn't match
+            obj = esc(ex)
+            return obj, ()
+        end
+        length(args) == 2 || error("Only 1- and 2-argument functions are supported")
+        sum(args_contain_under) == 1 || error("Only a single function argument can be the optic target")
+        if args_contain_under[1]
+            obj, frontoptic = parse_obj_optics(args[1])
+            optic = :(Base.Fix2($(esc(f)), $(esc(args[2]))))
+        elseif args_contain_under[2]
+            obj, frontoptic = parse_obj_optics(args[2])
+            optic = :(Base.Fix1($(esc(f)), $(esc(args[1]))))
+        end
     else
         obj = esc(ex)
         return obj, ()
--- a/test/test_delete.jl
+++ b/test/test_delete.jl
@@ -12,6 +12,11 @@ using StaticArrays
         @test delete( (1,2,3), last ) == (1, 2)
         @test @inferred(delete( (a=1, b=2, c=3), first ))== (b=2, c=3)
         @test @inferred(delete( SVector(1,2,3), last )) === SVector(1, 2)
+
+        l = @optic first(_, 2)
+        VERSION >= v"1.4" && @test l((1,2,3)) == [1,2]
+        @test delete((1,2,3), l) === (3,)
+
         let A = [1,2,3]
             @test delete(A, @optic(_[2])) == [1, 3]
             @test_throws Exception delete(A, @optic(_[2, 2]))
--- a/test/test_functionlenses.jl
+++ b/test/test_functionlenses.jl
@@ -94,4 +94,20 @@ end
 
 end
 
+@testset "custom binary function" begin
+    ↑(x, y) = x - y
+    Accessors.set(x, f::Base.Fix1{typeof(↑)}, y) = f.x - y
+    Accessors.set(x, f::Base.Fix2{typeof(↑)}, y) = f.x + y
+
+    x = 5
+    o1 = @optic 2 ↑ _
+    o2 = @optic _ ↑ 2
+    @test o1(x) == -3
+    @test set(x, o1, 10) == -8
+    @test o2(x) == 3
+    @test set(x, o2, 10) == 12
+    test_getset_laws(o1, x, 2, -3)
+    test_getset_laws(o2, x, 2, -3)
+end
+
 end # module


jw3126:
Thanks for @aplavin for exploring this direction, looks interesting!
>  If others agree this extension is a good idea, will add more methods and tests.

So the `first` example does not look very useful to me. But there might be other more practical cases. Do you have specific examples in mind, that might be practically useful?

aplavin:
A few usecases I have in mind:

Suppose there is a function that modifies the value located by an optic:
```
function f(xs, o)
    vals = o.(xs)
    ... compute new_vals ...
    set.(xs, o, new_vals)
end
```
The one can call it like `f([(a=1,), (a=2,), ...], @optic _.a)`. Want to operate on logarithms - also fine: `f([(a=1,), (a=2,), ...], @optic log(_.a))` (needs a `set` method for `log`).
But what if I need the function to operate on scaled values? `f([(a=1,), (a=2,), ...], @optic _.a * 1e10)` doesn't work without this PR. Same with units: `f([(a=1u"cm",), (a=2u"cm",), ...], @optic ustrip(u"m", _.a))`.

`first(_, n)` is also useful. Suppose the function `f` above works with 2d coordinates. Then `f([(coords=(1, 1),), (coords=(2, 1),), ...], @optic _.coords)` is fine. But how to apply it when data contain 3d coordinates? `@optic first(_.coords, 2)` solves this.

Some of these examples require defining corresponding `set` methods. I don't propose to define many function-lenses in this PR, and some (like `ustrip`) are impossible without depending on other packages. They are easy to define by downstream users, unlike the groundwork on optic parsing I propose here.

jw3126:
Thanks, @aplavin that makes sense to me. In code like `@set a*b = c`. What is returned `a_new` or `b_new`?
Also the `first` example can currently be expressed as `@optics _.coords[1:2]`.

aplavin:
> In code like @set a*b = c. What is returned a_new or b_new?

This shouldn't work at all, as before. My PR only concerns unambigous cases: the `@optic` macro, and only one of the function argument can contain `_`.

aplavin:
> Also the first example can currently be expressed as @optics _.coords[1:2].

Indeed, it can. I agree that `first` is less needed here.
Btw, I cannot make this work in current `Accessors.jl`: 
```
julia> x = (1, 2, 3)
julia> @set x[1:2] = (10, 20)
ERROR: MethodError: no method matching setindex(::Tuple{Int64, Int64, Int64}, ::Tuple{Int64, Int64}, ::UnitRange{Int64})
```
But that becomes orthogonal to this PR.

jw3126:
Ok, I am still not 100% convinced, but we can give it a shot. I also have quite some use cases, in the spirit of stripping units or adjusing scale. But usually, these cases are more complex and would not benefit from this PR. What I need from time to time is flatten parts of a nested struct into a rescaled vector that conforms some optimization API. In these cases and I define a custom optic directly. 
So my suggestion would be:
* We can merge this PR (and follow up PRs if required)
* You report back in a few months, whether you actually find this feature useful
Does that sound good to you?

aplavin:
> What I need from time to time is flatten parts of a nested struct into a rescaled vector that conforms some optimization API.

Oh, that's also useful! And I agree that this PR wouldn't really help here.
I sometimes do these btw:
image
So that calling `map(optics, array_of_objects)` returns a flat table with requested columns. But this is probably waaay to opinionated to be included in `Accessors.jl` itself.
<details><summary>Definitions</summary>
Sorry for formatting, code copied from a notebook.

```
# ╔═╡ 3de7af46-7375-44fa-857d-f730fcbae5a6
(nt::NamedTuple)(x) = map(f -> f(x), nt)

# ╔═╡ 6a115e94-9627-45ab-aed9-5950482faf45
begin
	flatten_composed(l::ComposedFunction) = (flatten_composed(l.inner)..., flatten_composed(l.outer)...)
	flatten_composed(l) = (l,)
end

# ╔═╡ 855f7f74-1328-46fa-bca6-e1667c61b32b
begin
	flat_key(l::ComposedFunction; nlast=1) = nlast == 1 ? flat_key(flatten_composed(l)[end]) :
		nlast == 2 ? Symbol(flat_key(flatten_composed(l)[end-1]), :(_), flat_key(flatten_composed(l)[end])) :
		nlast == 3 ? Symbol(flat_key(flatten_composed(l)[end-2]), :(_), flat_key(flatten_composed(l)[end-1]), :(_), flat_key(flatten_composed(l)[end])) :
		nlast == 4 ? Symbol(flat_key(flatten_composed(l)[end-3]), :(_), flat_key(flatten_composed(l)[end-2]), :(_), flat_key(flatten_composed(l)[end-1]), :(_), flat_key(flatten_composed(l)[end])) :
		nlast == 5 ? Symbol(flat_key(flatten_composed(l)[end-4]), :(_), flat_key(flatten_composed(l)[end-3]), :(_), flat_key(flatten_composed(l)[end-2]), :(_), flat_key(flatten_composed(l)[end-1]), :(_), flat_key(flatten_composed(l)[end])) :
		error("not implemented")
	flat_key(l::Accessors.PropertyLens{K}) where {K} = K
	flat_key(l::Accessors.IndexLens) = only(l.indices)
	flat_key(l::Function) = nameof(l)
end

# ╔═╡ 210aeab0-5ec9-48fc-8e97-bd325e51e853
function w_flat_keys(ls)
	res = flat_key.(ls)
	i = 1
	while !allunique(res)
		@show res i
		i += 1
		dct = @p group(res, ls) |>
			mapmany() do gr
				length(gr) == 1 && return []
				keys = flat_key.(gr; nlast=i)
				gr .=> keys
			end |>
			Dict
		res = get.(Ref(dct), ls, res)
	end
	return res .=> ls
end

# ╔═╡ 06eb1166-3697-49ce-b5c6-e04df608f5cb
begin
	to_flat_nt(x::NamedTuple) = x
	to_flat_nt(x::Tuple) = (;w_flat_keys(x)...)
	to_flat_nt(x::AbstractVector) = to_flat_nt(Tuple(x))
	to_flat_nt(x) = to_flat_nt((x,))
end
```

</details>

aplavin:
> So my suggestion would be:
> 
> We can merge this PR (and follow up PRs if required)
> You report back in a few months, whether you actually find this feature useful
> Does that sound good to you?

That totally works for me!

aplavin:
>  In these cases and I define a custom optic directly.

Do you have any neat or simple examples? I'm not really experienced with custom optics, outside of defining `set` for existing functions.

jw3126:
> > In these cases and I define a custom optic directly.
> 
> Do you have any neat or simple examples? I'm not really experienced with custom optics, outside of defining `set` for existing functions.

My code looks close to your code, with one addition. I often need constraint optimization and I often incorportate the constraints by transforming my variables. For instance, say I want to minimize `f(x)` where `x` is in the interval `[lo,hi]`, but my solver only supports unconstraint optimization. Then one can define an optic like:
```julia
struct Stretch{T}; lo::T; hi::T; end
(o::Stretch)(x_bounded) = # use bijection from [lo, hi] to IR
Accessors.set(_, optic::Stretch, y_unbound) = # use inverse bijection

function optimize(f, obj0, lens)
    v0 = len(obj0)
    function _f(v)
         obj = set(obj0, lens, v)
         f(obj)
    end
    _sol = LibThatExpectsVectors.optimize(f0, v0)
    set(obj0, lens, _sol)
end
```
Then I build more complicated optics from such building blocks, similar to your code above.

aplavin:
First I thought that it's better to add more lenses in the same PR, but now I believe it's best to merge this as is. There is one example and a test, confirming that the new machinery works fine.

I think I'm going to propose new lenses separately, so that potential discussions of what to include don't affect the two-argument function parsing itself.
Mostly, I think about arithmetic, introduce `set` methods for
- invertible single-argument functions in `Base`, like `exp`, `log`, `sqrt`, `angle`,
- two-argument functions (based on this PR), like `+`, `*`, `^`.

aplavin:
>  This happens because @set a*b = c is parsed as obj = a*b, optic=identity:
> I consider it a bug and if your PR incidentally turns it into an error, that's fine.

This PR doesn't do anything against the case you mentioned. Furthermore, the current behavior is not **that** unexpected sometimes, and even tested in the testsuite: https://github.com/JuliaObjects/Accessors.jl/blob/fb8fb821a7aaf7b6d7361d62f55e5a233eaa1f5d/test/test_core.jl#L123
Still not sure if I like this or not, but it's not always a bug.

jw3126:
> This PR doesn't do anything against the case you mentioned. Furthermore, the current behavior is not **that** unexpected sometimes, and even tested in the testsuite:

Yes it is not the job of the PR to address this, sorry if I made that impression. It was just that if this PR broke that behavior, that would be ok to me.

> Still not sure if I like this or not, but it's not always a bug.

Yes this was intended in ancient times, where we had no function lenses. But now I think it is a misfeature and I am happy to sacrifice it.


Can you add a test, that simulates a user defining `mybinop` and overaloading the required methods on Fix1/Fix2 ?

aplavin:
> Yes this was intended in ancient times, where we had no function lenses. But now I think it is a misfeature and I am happy to sacrifice it.

While we are at it, what do you think of making `@set $(v[i]).a = 5` equivalent to `vi = v[i]; @set vi.a`? That is, expression in `$`  treated as the `@set` target.
Then instead of `@set T(1,2).a = 2` one would write `@set $(T(1,2)).a = 2`

> Can you add a test, that simulates a user defining mybinop and overaloading the required methods on Fix1/Fix2 ?

Added!

jw3126 (approved):
Thanks a lot LGTM. Shall I merge?

PR #48, opened 2022-04-24
Document direct constructors, add convenience form.

### LINKED ISSUE #47: expose API to build lenses
Setfield.jl allowed interpolation with `$`, but that of course has some corner cases so I understand why it is not allowed in `@optic`. However, it would be nice to be able to build lenses with functions, eg for a

```julia
(a = 1:2, b = 1:3)
```
build the equivalent of `[@optic(_.a[1]), @optic(_.a[2]), @optic(_.b[1]), @optic(_.b[2]), @optic(_.b[3])]` with a loop or similar, eg with a syntax like `Property(:a) ∘ Index(1)` etc.

tpapp:
- add convenience constructors for `IndexLens` and `PropertyLens`
- export and document both
- explain direct lens construction in the manual

Fixes #47.

### DIFF
--- a/docs/src/lenses.md
+++ b/docs/src/lenses.md
@@ -38,6 +38,9 @@ julia> v = (a = 1:3, )
 julia> l = opcompose(PropertyLens(:a), IndexLens(1))
 (@optic _[1]) ∘ (@optic _.a)
 
+julia> l ≡ @optic _.a[1]   # equivalent to macro form
+true
+
 julia> l(v)
 1
 
--- a/docs/src/lenses.md
+++ b/docs/src/lenses.md
@@ -3,6 +3,7 @@
 Accessors.jl is build around so called lenses. A Lens allows to access or replace deeply nested parts of complicated objects.
 
 # Example
+
 ```jldoctest
 julia> using Accessors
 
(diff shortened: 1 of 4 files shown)


test/test_optics.jl
```diff
@@ -82,4 +82,9 @@ end
     @inferred modify(x -> 0, arr, @optic _ |> Elements() |> If(iseven))
 end
 
+@testset "convenience constructors" begin
+    @test IndexLens(1, 2, 3) == IndexLens((1, 2, 3))
```
jw3126, line 86:
```suggestion
    @test IndexLens(1, 2, 3) === IndexLens((1, 2, 3))
```

test/test_optics.jl
```diff
@@ -82,4 +82,9 @@ end
...
+    @test PropertyLens(:a) == PropertyLens{:a}()
```
jw3126, line 87:
```suggestion
    @test PropertyLens(:a) === PropertyLens{:a}()
```

docs/src/lenses.md
```diff
@@ -26,6 +27,24 @@ julia> modify(lowercase, obj, lens)
...
 
+Lenses can also be constructed directly and composed with [`opcompose`](@ref), `⨟`, or `∘` (note reverse order).
+
+```jldoctest
+julia> using Accessors
+
+julia> v = (a = 1:3, )
+(a = 1:3,)
+
+julia> l = opcompose(PropertyLens(:a), IndexLens(1))
```
jw3126, line 38: Maybe add a line like:
```
julia> opcompose(PropertyLens(:a), IndexLens(1)) === @optic _.a[1]
true
```
To demonstrate equivalence of the explicit form with the macro.

jw3126 (approved):
Overall looks good, thanks a lot! I added some minor suggestions, see above.

docs/src/lenses.md
tpapp: done, thanks for all the suggestions!

jw3126:
Awesome if you bump the patch version, I will tag a release!

tpapp:
done

jw3126:
Thanks again for the PR!

PR #53, opened 2022-06-07
add mapview() support

aplavin:
I really like this feature, and how simple it was to implement!
Transparently set values on mapped array views (from `SplitApplyCombine.jl`) when the mapping function is an optic.
The simplest example from tests:
```julia
X = [(a=1, b=2), (a=3, b=4)]
Y = mapview(@optic(_.b), X)
@test Y == [2, 4]
Y[2] = 100
@test X == [(a=1, b=2), (a=3, b=100)]
```

Haven't noticed before that `Accessors.jl` already uses `Requires`, which makes this addition possible without extra dependencies.

### DIFF
--- a/src/mapview.jl
+++ b/src/mapview.jl
@@ -1,0 +1,3 @@
+import .SplitApplyCombine: MappedArray
+
+Base.setindex!(ma::MappedArray, val, ix) = parent(ma)[ix] = set(parent(ma)[ix], ma.f, val)
--- a/Project.toml
+++ b/Project.toml
@@ -31,9 +31,10 @@ Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
 InteractiveUtils = "b77e0a4c-d291-57a0-90e8-8db25a27a240"
 PerformanceTestTools = "dc46b164-d16f-48ec-a853-60448fc869fe"
 QuickTypes = "ae2dfa86-617c-530c-b392-ef20fdad97bb"
+SplitApplyCombine = "03a91e81-4c3e-53e1-a0a4-9c0c8f19dd66"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 StaticNumbers = "c5e4b96a-f99f-5557-8ed2-dc63ef9b5131"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [targets]
-test = ["Test", "Documenter", "PerformanceTestTools", "QuickTypes", "StaticArrays", "BenchmarkTools", "InteractiveUtils", "StaticNumbers"]
+test = ["Test", "Documenter", "PerformanceTestTools", "QuickTypes", "SplitApplyCombine", "StaticArrays", "BenchmarkTools", "InteractiveUtils", "StaticNumbers"]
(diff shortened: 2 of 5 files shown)


src/mapview.jl
```diff
@@ -0,0 +1,3 @@
+import .SplitApplyCombine: MappedArray
+
+Base.setindex!(ma::MappedArray, val, ix) = parent(ma)[ix] = set(parent(ma)[ix], ma.f, val)
```
jw3126, line 3: The feature is very neat, but this is [type piracy](https://docs.julialang.org/en/v1/manual/style-guide/#Avoid-type-piracy). One possibility would be to create a zero dependencies package `AccessorsBase` that just contains something like
```julia
function set end
```
and ask @andyferris if we can add the `Base.setindex!` overload at SplitApplyCombine.

aplavin: That would also be piracy: all these functions, at least https://github.com/JuliaObjects/Accessors.jl/blob/master/src/functionlenses.jl

aplavin: Maybe, "pirating" a method that isn't defined, and won't feasibly be defined in the original library is fine? As I understand, the main concern is that piracy can change the behavior of unrelated code - here it's hard to imagine such a case. SAC.jl cannot really define setindex for an arbitrary mappedarray.

jw3126: > That would also be piracy: all these methods, at least https://github.com/JuliaObjects/Accessors.jl/blob/master/src/functionlenses.jl

That is fine by the 

> Sometimes, coupled packages may engage in type piracy to separate features from definitions

clause in the manual. If we overloaded `UnsupectingPackage.set` that would be a problem.

> Maybe, "pirating" a method that isn't defined, and won't feasibly be defined in the original library is fine? 

Fair enough. @andyferris do you think we can do this overload here?

jw3126: > Maybe, "pirating" a method that isn't defined, and won't feasibly be defined in the original library is fine? 

What can happen is that some other people think the same and provide overloads for `Base.setindex!(::MappedArray...)` Then a third package might have `Accessors` and the other package as transitive dependencies and get weird bugs.

aplavin: This PR doesn't introduce any new dependency, `Requires` is already used in `Accessors`.

rafaqz: Ok I just saw the growinh Project.toml. Thats what people will look at to check dependencies.

Mostly everyone is moving away from Requires.jl because its a hack with serious ttfx overheads - it cant precompile.

jw3126: Accessors.jl getting heavy is a good point @rafaqz and getting rid of Requires.jl is also on my wishlist. Currently, we only need it to avoid depending on StaticArrays.jl . 
Do you have a suggestion where stuff like this PR could live?

rafaqz: Well we really conditional dependencies: https://github.com/JuliaLang/Pkg.jl/issues/1285

What ArrayInterface.jl is doing and we're currently doing with GeoInterface.jl is making subpackages in the same repository, names as the combination of packages. Like ArrayInterfaceOffsets.jl or GeoInterfaceTables.jl.

Then other packages and users can choose their dependencies. Its pretty clunky but the best we have currently. 

(also sorry I meant to put my initial comment in the main thread I commented from a phone)

aplavin: I'm also curious how the StaticArrays support in Accessors would look without Requires. Should this https://github.com/JuliaObjects/Accessors.jl/blob/master/src/staticarrays.jl file be put into a separate package that depends on both? The major issue then is that users wouldn't install it and will think that Accessors just doesn't support SVectors.

PR #55, opened 2022-06-11
support explicit specification of the optic target

aplavin:
The main common usecase is using `@(re)set` with an array element, as in `@set $(x[i]).a.b = 123` and `@reset $(x[i]).a.b = 123`.
Currently, these require a temporary variable:
```
old = x[i]
new = @set old.a.b = 123
```
and
```
old = x[i]
x[i] = @set old.a.b = 123
```

### DIFF
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -45,6 +45,8 @@ end
     nt = (a=1,)
     @reset nt.a = 5
     @test nt === (a=5,)
+
+    @test_throws Exception eval(:(@reset func(x, y) = 100))
 end
 
 @testset "@set" begin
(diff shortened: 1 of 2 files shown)


test/test_core.jl
```diff
@@ -136,6 +136,21 @@ end
...
+    x = [1, 2, 3]
+    x_orig = x
+    @test (@set $(x)[2] = 100) == [1, 100, 3]
+    @test (@set $(x[2]) = 100) == 100
+    @test (@set $(x)[2] + 2 = 100) == [1, 98, 3]  # impossible without $
+    @test (@set $(x[2]) + 2 = 100) == 98  # impossible without $
+    @test x_orig === x == [1, 2, 3]
+
+    @test (@reset $(x[2]) = 100) == 100
+    @test x_orig === x == [1, 100, 3]
```
jw3126, line 149: This makes sense once you think about the macro expansion, but is still a gotcha I think. Maybe we should forbid interpolation in `@reset`?

jw3126: Reading your post again, I now realize that array mutation is very much intended.

jw3126: Right now, Accessors.jl never mutates anything. This is a nice and simple rule. This feature would break that rule and also adds more complex syntax. This is of course no issue for Accessors.jl power users like you. 
But most people are not Accessors.jl power users, they might just encounter `@reset $(x[i]).a.b = 123` while reading some elses code.
For these people, this feature adds friction without much benefit. They might need to search through the Accessors.jl docs/source and waste time debugging a strange mutation.

So overall, I would prefer to avoid this feature in favor of simplicity. Are you ok with that @aplavin ?

jw3126: I have not heard of it in other ecosystems, but I agree mutating lenses can be useful sometimes.
There is a toy implementation here:
https://juliaobjects.github.io/Accessors.jl/stable/examples/custom_macros/

We can try to make that implementation more solid and add it to the package. 
I think instead of 
```julia
function Accessors.set(o, l::Lens!{PropertyLens{prop}}, val) where {prop}
    setproperty!(o, prop, val)
    o
end
```
We probably want to add a `set!` function for easier overload:
```julia
set(o, l::Lens!, val) = set!(o, l.pure, val) # Accessors.jl
Accessors.set!(o,::PropertyLens, val) = ... # Accessors.jl
Accessors.set!(o,::MyLens, val) = ... # user code
```
With this one can do already
```julia
@set! a.b.c.d = e
```
which would mutate `c`.
We might also want some syntactic sugar to control where the mutation happens in 
```julia
a.b.c.d = e 
``` 
One can mutate `b` with
```julia
@set a |> @optic!(_.b.c) |> @optic(_.d) = e 
```
but this is confusing and verbose.

jw3126:
Agreed, this is nice.

PR #56, opened 2022-06-28
run doctests only for julia >= 1.7

jw3126:


test/runtests.jl
```diff
@@ -14,13 +14,4 @@ include("test_setmacro.jl")
 include("test_setindex.jl")
 include("test_functionlenses.jl")
 
-using Documenter: doctest
```
jw3126, line 17: I think running doctests here just causes more headaches than it adds values. So let's remove them. They are run in the docs build anyway.

PR #57, opened 2022-07-03
support setting fields

aplavin:
needs the corresponding ConstructionBase branch

### DIFF
--- a/src/functionlenses.jl
+++ b/src/functionlenses.jl
@@ -18,6 +18,9 @@ function set(obj, ::typeof(only), val)
     set(obj, first, val)
 end
 
+set(obj, o::typeof(getfields), val) = constructorof(typeof(obj))(val...)
+set(obj, o::Base.Fix2{typeof(getfield)}, val) = @set getfields(obj)[o.x] = val
+
 ################################################################################
 ##### eltype
 ################################################################################
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -3,7 +3,7 @@ using Test
 using Accessors
 using Accessors: test_getset_laws, test_modify_law
 using Accessors: compose, get_update_op
-using ConstructionBase: ConstructionBase
+using ConstructionBase: ConstructionBase, getfields
 using StaticNumbers: StaticNumbers, static
 if !isdefined(Base, :only)
     using Accessors: only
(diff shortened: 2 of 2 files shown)


src/functionlenses.jl
```diff
@@ -18,6 +18,13 @@ function set(obj, ::typeof(only), val)
     set(obj, first, val)
 end
 
+function set(obj, o::Base.Fix2{typeof(getfield)}, val)
+    nt = getfields(obj)
+    nt_new = @set nt[o.x] = val
+    args = Tuple(nt_new) # old julia inference prefers if we wrap in Tuple
```
jw3126, line 24: I think this wrapping is needed only for julia < 1.3 which Accessors.jl does not support anyway.

test/test_core.jl
```diff
@@ -378,29 +382,41 @@ end
...
 @testset "setproperties overloading" begin
-    o = CustomProperties("A", "B")
+    o = CustomProperties("A", "B", "C")
     o2 = @set o.a = :A
-    @test o2 == CustomProperties(:A, "B")
+    @test o2 == CustomProperties(:A, "B", "C")
     o3 = @set o.b = :B
-    @test o3 == CustomProperties("A", :B)
+    @test o3 == CustomProperties("A", :B, "C")
+
```
jw3126, line 413: Some inference tests would be good.

jw3126: Somethink like
```julia
function set_field_a(obj, val)
    @set getfield(obj,:a) = val
end
@inferred set_field_a(obj, val)
```

jw3126: That is a good point. `getfield(_, :a)` will store the `:a` as a runtime symbol. Which is bad for "long range" inference like here. 
Maybe a better approach is to introduce `FieldLens` analogous to `PropertyLens`. It can store the field as part of the type and enable long range inference.

I think manipulating fields is a rare enough use case to justify ugly syntax like `@set FieldLens(:a)(obj) = val`.

jw3126: I think the analogy with `IndexLens` is flawed. First of all `getindex` is usually type stable (think of Vectors, Dicts etc.), while `getfield` and `getproperty` are rarely type stable.
So it is a sane default for `IndexLens` to store the index as a field and a sane default for `PropertyLens` to store it in the type. If the default is not ideal, say because you want to manipulate Tuples over long range, you can request that:
```julia
using StaticNumbers
lens = @optic _[static(2)] # Tuple access will be type stable with this
```
So I think the analogy between field access  and property access is much closer than between field access and index access. Therefore we should mimic PropertyLens not IndexLens here.

PR #60, opened 2022-07-08
Drop Requires, depend on StaticArraysCore

cscherrer:
Because of method invalidations, Requires.jl can cause package load times to be much longer. Accessors currently uses Requires for StaticArrays.jl, which is fairly heavy-weight. But that package has recently been refactored to add a StaticArraysCore.jl, which is very light.

This little PR drops use of Requires in favor of a dependency on StaticArraysCore.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -23,7 +23,7 @@ MacroTools = "0.4.4, 0.5"
 StaticArraysCore = "1"
 StaticNumbers = "0.3"
 Unitful = "1"
-julia = "1.3"
+julia = "1.6"
 
 [extras]
 BenchmarkTools = "6e4b80f9-dd63-53aa-95a3-0cdb28fa8baf"
--- a/Project.toml
+++ b/Project.toml
@@ -11,7 +11,7 @@ Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
-Requires = "ae029012-a4dd-5104-9daa-d747884805df"
+StaticArraysCore = "1e83bf80-4336-4d27-bf5d-d5a4f845583c"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [compat]
@@ -20,7 +20,7 @@ CompositionsBase = "0.1"
 ConstructionBase = "1.2"
 InverseFunctions = "0.1.5"
 MacroTools = "0.4.4, 0.5"
-Requires = "0.5, 1.0"
+StaticArraysCore = "1"
 StaticNumbers = "0.3"
 julia = "1.3"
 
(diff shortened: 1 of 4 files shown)


jw3126:
Great I would love to get rid of Requires. The CI fails are StaticArrays related. Do the tests pass locally for you? Does it depend on the StaticArrays version?

cscherrer:
Failing for me too. Not sure why, could be a StaticArrays version thing

jw3126:
Locally on `master` tests pass for me with `StaticArrays 1.5.0`.

Project.toml
```diff
@@ -20,7 +20,7 @@ CompositionsBase = "0.1"
 ConstructionBase = "1.2"
 InverseFunctions = "0.1.5"
 MacroTools = "0.4.4, 0.5"
-Requires = "0.5, 1.0"
+StaticArraysCore = "1"
 StaticNumbers = "0.3"
 julia = "1.3"
```
oschulz, line 25:
```suggestion
julia = "1.6"
```

oschulz: I think Accessors needs to require Julia v1.6 for StaticArraysCore.

cscherrer:
My local tests are still failing on Julia 1.8. Do we need to constrain versions in the test environment? I think that can be done by putting another `Project.toml` in `/test`, not sure if there's another way

thomvet:
I just did literally the same Pull Request on Setfield.jl (https://github.com/jw3126/Setfield.jl/pull/174) where tests pass locally (let's see what happens on CI). 

If it works there, maybe it helps someone with more insight into the differences between the packages to identify the cause here?

rafaqz:
Requires blocks can invalidate code in other packages, and being at the bottom of a package stack that adds up. 

Its really which lines, not how many lines, that matters.

cscherrer:
> As I said in a comment above, you just removed `include("staticarrays.jl")`. Add it back, to all other includes.

I don't see any earlier comments from you, but this is a great point. Now the problem is these lines:
```julia
@inline delete(obj::StaticArraysCore.SVector, l::IndexLens) = StaticArraysCore.deleteat(obj, only(l.indices))
@inline insert(obj::StaticArraysCore.SVector, l::IndexLens, val) = StaticArraysCore.insert(obj, only(l.indices), val)
```

`deleteat` and `insert` are in StaticArrays, not StaticArraysCore. So this is still broken. I think the solution is to change StaticArraysCore to declare these functions. OTOH that would lead to code invalidation when loading StaticArrays. Any better ideas?

oschulz:
>  Now the problem is these lines:
> ...
> ... change StaticArraysCore to declare these functions.

Yes, we could ask for that - maybe make PRs to StaticArrays and -Core directly. Just declaring the functions in StaticArraysCore would be non-controversial (I would hope).

cscherrer:
> Yes, we could ask for that - maybe make PRs to StaticArrays and -Core directly. Just declaring the functions in StaticArraysCore would be non-controversial (I would hope).

Seems to be the case:
https://github.com/JuliaArrays/StaticArraysCore.jl/issues/11

> Github marks my older comment as "pending", maybe it's only visible to me then? But I don't see any button to actually submit it:

When you add a comment, you can either add it as a one-off or start a review. You must have done the latter. Then you need to click to submit. There should be a button at the top right, "Complete review" or "submit review" or something.

> As for Requires overhead: so, unfortunately no simple example anywhere?

If you do `@time_imports using MeasureTheory` before and after this PR:

**Before**:
```
    675.1 ms  Accessors 79.25% compilation time (48% recompilation)
```
**After**:
    123.9 ms  Accessors
```

jw3126:
I close this in favor of #78 Feel free to reopen, if I miss something.

PR #61, opened 2022-07-09
fix setting real and abs

aplavin:
now they work properly with other number types, not just Real and Complex

### DIFF
--- a/src/functionlenses.jl
+++ b/src/functionlenses.jl
@@ -48,11 +48,10 @@ delete(path, ::typeof(dirname)) = basename(path)
 ################################################################################
 ##### math
 ################################################################################
-set(x::Real, ::typeof(real), y) = y
-set(x,       ::typeof(real), y) = y + im*imag(x)
-set(x,       ::typeof(imag), y) = real(x) + im*y
-set(x,      ::typeof(angle), y) = abs(x) * cis(y)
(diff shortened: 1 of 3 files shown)


src/functionlenses.jl
```diff
@@ -48,11 +48,12 @@ delete(path, ::typeof(dirname)) = basename(path)
...
 ##### math
 ################################################################################
-set(x::Real, ::typeof(real), y) = y
-set(x,       ::typeof(real), y) = y + im*imag(x)
-set(x,       ::typeof(imag), y) = real(x) + im*y
-set(x,      ::typeof(angle), y) = abs(x) * cis(y)
-set(x,        ::typeof(abs), y) = y >= 0 ? y * sign(x) : throw(DomainError(y, "cannot set abs($x) to $y"))
+set(x::Real,    ::typeof(real), y) = y
+set(x::Complex, ::typeof(real), y) = y + im*imag(x)
+set(x         , ::typeof(real), y) = x - real(x) + y
```
jw3126, line 53: I think this method is problematic when `x = MyNumber(Inf)`.

jw3126: What would be a `typeof(x)` that gets handled by this method?

jw3126 (approved):
LGTM

Issue #63, opened 2022-07-13
get/set all values referred by an optic


aplavin:
Some existing optics, such as `Elements()`, `If()` and `Recursive()`, can only be `modify`ed - one cannot retrieve the values specified by them. I understand that it's not obvious how to implement `optic(obj)` for them and even what type it should return. However, from time to time I try to reach for this functionality: either to specifically extract the corresponding values, or just to see what values `modify` would be called with.

Do you think it's possible to design value extraction for those optics in a consistent manner?

jw3126:
A quick and dirty way to do this is:
```julia
out = Any[]
modify(obj, optic) do val
    push!(out, val)
    f(val)
end
```
@rafaqz has done a lot of work in this direction, see https://github.com/JuliaObjects/Accessors.jl/pull/23
Ultimately we could not find a performant way to implement this, that's why the PR is not merged.

jw3126:
Okay, so is your question whether one can implement `optic(obj)` for these such that the lens laws are satisfied?

I think that is not really possible. For instance:
```julia
julia> using Accessors

julia> obj = [1,2,3]
3-element Vector{Int64}:
 1
 2
 3

julia> optic = Elements()
Elements()

julia> modify(identity, obj, optic)
3-element Vector{Int64}:
 1
 2
 3

julia> set(obj, optic, 1)
3-element Vector{Int64}:
 1
 1
 1
```
That means you can only ever construct constant vectors using `set`, `Elements()`, but you can produce arbitrary vectors using `modify`, `Elements()`. So there is no hope to decompose `modify` into `set` and "get".

Mathematically there are many flavors of optics and these can have different APIs and laws. Lens is one with lots of structure and laws. `Elements()` only satisfies the functor laws and does not have a "get" operation:
* `@assert modify(identity, optic, obj) == obj`
* `@assert modify(f ∘g, optic, obj) == modify(f, optic, modify(g, optic, obj))`

rafaqz:
> #23 looks a bit over my head - not even talking about implementation, I don't fully understand the interface and usecases.
> Extracting values with `Elements` and `If` is probably much easier to implement performantly than very general #23 queries. A major issue is the interface: it should be consistent with
> https://github.com/JuliaObjects/Accessors.jl/blob/fa0a446dab58fc6bcfc5e97413c441da04ec3e02/src/testing.jl#L18-L24

The use case is reconstructing arbitrary immutable objects with new/modified field values.

[...]

PR #64, opened 2022-07-16
add getall(obj, optic)

aplavin:
Following #63 discussion.
Even `getall` for long optic chains doesn't infer: see the broken test. Tried adding `@inline` everywhere, but it doesn't change anything.

### DIFF
--- a/src/getsetall.jl
+++ b/src/getsetall.jl
@@ -1,4 +1,30 @@
-getall(obj, ::Elements) = obj |> values
+"""
+    getall(obj, optic)
+
+Extract all parts of `obj` that are selected by `optic`.
+Returns a flat `Tuple` of values, or an `AbstractVector` if the selected parts contain arrays.
+This function is experimental and we might change the precise output container in future.
+
+
+```jldoctest
+julia> using Accessors
+
+julia> obj = (a=1, b=(2, 3));
+
+julia> getall(obj, @optic _.a)
+(1,)
+
+julia> getall(obj, @optic _ |> Elements() |> last)
+(1, 3)
+```
+"""
+function getall end
+
+getall(obj::Union{Tuple, AbstractVector}, ::Elements) = obj
+getall(obj::Union{NamedTuple}, ::Elements) = values(obj)
+getall(obj::AbstractArray, ::Elements) = vec(obj)
+getall(obj::Number, ::Elements) = (obj,)
+getall(obj::AbstractString, ::Elements) = collect(obj)
 getall(obj, ::Properties) = getproperties(obj) |> values
 getall(obj, o::If) = o.modify_condition(obj) ? (obj,) : ()
 getall(obj, f) = (f(obj),)
@@ -7,13 +33,13 @@ getall(obj, f) = (f(obj),)
 # A recursive implementation of getall doesn't actually infer,
 # see https://github.com/JuliaObjects/Accessors.jl/pull/64.
 # Instead, we need to generate unrolled code explicitly.
-function getall(obj, optic::CO) where {CO <: ComposedFunction}
+function getall(obj, optic::ComposedFunction)
     N = length(decompose(optic))
     _GetAll{N}()(obj, optic)
 end
 
 struct _GetAll{N} end
-(::_GetAll{N})(_) where {N} = error("Too many chained optics: $N is not supported for now.")
+(::_GetAll{N})(_) where {N} = error("Too many chained optics: $N is not supported for now. See also https://github.com/JuliaObjects/Accessors.jl/pull/64.")
 
 _concat(a::Tuple, b::Tuple) = (a..., b...)
 _concat(a::Tuple, b::AbstractVector) = vcat(collect(a), b)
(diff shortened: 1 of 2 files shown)


test/test_getsetall.jl
```diff
@@ -0,0 +1,21 @@
...
+    obj = (a=1, b=2.0, c='3')
+    @test @inferred(getall(obj, @optic(_.a))) === (1,)
+    @test @inferred(getall(obj, @optic(_ |> Properties()))) === (1, 2.0, '3')
+    @test @inferred(getall(obj, @optic(_ |> Elements()))) === (1, 2.0, '3')
+    
+    obj = (a=1, b=((c=3, d=4), (c=5, d=6)))
+    @test @inferred(getall(obj, @optic(_.b |> Elements() |> _.c))) === (3, 5)
+    @test @inferred(getall(obj, @optic(_.b |> Elements() |> Properties()))) === (3, 4, 5, 6)
+    @test getall(obj, @optic(_.b |> Elements() |> Properties() |> _ * 3)) === (9, 12, 15, 18)
+    @test_broken (@inferred(getall(obj, @optic(_.b |> Elements() |> Properties() |> _ * 3))); true)
```
jw3126, line 15: Can you add a few more tests for when inference works and when it is broken? In particular, I am curious if it is a problem that we have a function lens here. As I understand the compiler has some heuristics not to specialize too much on arguments that are functions.

rafaqz:
Hard coded nesting... why not use a generated function for that?

src/optics.jl
```diff
@@ -465,3 +465,145 @@ function make_salt(s64::UInt64)::UInt
...
+
+    reduce(_concat,
+        map(getall(obj, f1)) do obj
+            reduce(_concat,
+                map(getall(obj, f2)) do obj
+                    reduce(_concat,
+                        map(getall(obj, f3)) do obj
+                            reduce(_concat,
+                                map(getall(obj, f4)) do obj
+                                    getall(obj, f6)
```
jw3126, line 533:
```suggestion
                                    getall(obj, f5)
```

jw3126: Probably also need to cover each of these in tests.

rafaqz:
Why not? All options are bad though lol.

But do look at `@assume_effects` in 1.8, it should remove the need for all of this.

rafaqz:
I didnt look that hard I thought you were brute forcing it so there were no closures. 

But again, `@assume_effects` should fix the type stability in 1.8. I would lean towards that rather than any kind of code generation.

rafaqz:
You can always do recirsive code generation like Flatten.jl ;)

src/getsetall.jl
```diff
@@ -0,0 +1,50 @@
...
+    end
+
+    :(function (::_GetAll{$N})(obj, optic)
+        ($(syms...),) = deopcompose(optic)
+        $expr
+    end) |> esc
+end
+
+for i in 2:10
+    @eval @_generate_getall $i
```
jw3126, line 49: Why introduce a macro here? Why not `eval(generate_getall(i))`?

src/getsetall.jl
```diff
@@ -0,0 +1,50 @@
...
+# A recursive implementation of getall doesn't actually infer,
+# see https://github.com/JuliaObjects/Accessors.jl/pull/64.
+# Instead, we need to generate unrolled code explicitly.
+function getall(obj, optic::CO) where {CO <: ComposedFunction}
+    N = length(decompose(optic))
+    _GetAll{N}()(obj, optic)
+end
+
+struct _GetAll{N} end
+(::_GetAll{N})(_) where {N} = error("Too many chained optics: $N is not supported for now.")
```
jw3126, line 16:
```suggestion
(::_GetAll{N})(_) where {N} = error("Too many chained optics: $N is not supported for now. See also https://github.com/JuliaObjects/Accessors.jl/pull/64")
```

src/getsetall.jl
```diff
@@ -0,0 +1,50 @@
...
+        )
+    end
+
+    :(function (::_GetAll{$N})(obj, optic)
+        ($(syms...),) = deopcompose(optic)
+        $expr
+    end) |> esc
+end
+
+for i in 2:10
```
jw3126, line 48: Can inference handle chains of length 10 with this? Maybe add a test?

jw3126 (commented):
Thanks a lot, @aplavin. I like it. I mean it is ugly, but probably it is the best one can do with the current compiler. Needs some docs and I had some minor comments above. Otherwise LGTM.

src/getsetall.jl
```diff
@@ -0,0 +1,70 @@
+"""
+    getall(obj, optic)
+
+Extract all parts of `obj` that are selected by `optic`.
+Returns a flat tuple of values, or a vector if the selected parts contain vectors.
```
jw3126, line 5:
```suggestion
Returns a flat `Tuple` of values, or a `Vector` if the selected parts contain vectors.
This function is experimental and we might change the precise output container in future.
```
For instance, we might want to return SVectors in certain situtations.

jw3126:
> I noticed an issue with `getall(Elements)` and some types, such as numbers and strings:
> 
> ```julia
> julia> getall(1, Elements())
> 1  # should be (1,)
> 
> julia> getall("abc", Elements())
> "abc"  # should be ['a', 'b', 'c']
> ```
> 
> These are easily fixable, but I'm not sure of the best way to do that. Just special-case numbers and strings? Is there anything else?

Mhh good point. Is it a feature or a bug? In a way it is consistent:
```julia

julia> map(identity, 1)
1

julia> map(identity, "asdf")
"asdf"
```
You think 
> ```julia
> julia> getall(1, Elements())
> 1  # should be (1,)
> 
> julia> getall("abc", Elements())
> "abc"  # should be ['a', 'b', 'c']
> ```
Will cause trouble in practice?
I think we can demand the following from `out = getall(obj, optic)`:
* Is an iterator that supports length, eachindex, getindex
* Has the same elements as
```julia
out2 = []
modify(obj, optic) do val
   push!(out2, val)
end
```
Maybe that is all we need?

jw3126 (approved):
LGTM nice! Do you plan to do `setall` as well?

PR #65, opened 2022-07-16
simplify and generalize mapproperties()

aplavin:
noticed that get/setproperties work for Tuples, but mapproperties doesn't


jw3126 (approved):
LGTM thanks!

PR #68, opened 2022-07-21
add setall()

aplavin:
Copy-pasted till depth=4, seems to infer fine. The plan is to make the same kind of macro as with `getall` in the end, after all other questions are resolved.

The implementation doesn't seem very clean, though - unlike `getall`. Improvement suggestions are welcome!
Also, it's likely not the most efficient: `getall` is called more times than strictly needed. Would be great to remove extra calls, the complication is that we don't want `setall` to recurse: currently, it only calls `setall` with individual (not composed) optics.

### DIFF
--- a/src/getsetall.jl
+++ b/src/getsetall.jl
@@ -3,7 +3,10 @@
 
 Extract all parts of `obj` that are selected by `optic`.
 Returns a flat `Tuple` of values, or an `AbstractVector` if the selected parts contain arrays.
-This function is experimental and we might change the precise output container in future.
+
+This function is experimental and we might change the precise output container in the future.
+
+See also [`setall`](@ref).
 
 
 ```jldoctest
@@ -19,6 +22,32 @@ julia> getall(obj, @optic _ |> Elements() |> last)
 ```
 """
 function getall end
+
+"""
+    setall(obj, optic, values)
+
+Replace a part of `obj` that is selected by `optic` with `values`.
+The `values` collection should have the same number of elements as selected by `optic`.
+
+This function is experimental and might change in the future.
+
+See also [`getall`](@ref), [`set`](@ref). The former is dual to `setall`:
+
+```jldoctest
+julia> using Accessors
+
+julia> obj = (a=1, b=(2, 3));
+
+julia> optic = @optic _ |> Elements() |> last;
+
+julia> getall(obj, optic)
+(1, 3)
+
+julia> setall(obj, optic, (4, 5))
+(a = 4, b = (2, 5))
+```
+"""
+function setall end
 
 # implementations for individual noncomposite optics
 


jw3126 (commented):
Thanks a lot for giving this a shot!

For me personally, the really useful cases would be, where `getall` and `setall` work with `Vector`, while your PRs focus on the `Tuple` case. 

Implementation wise, I might have tried to use less `getall`. I think in many places
where you call `getall` you really only need to know `length(getall(...))` or maybe `Val(length(getall(...)))`. So maybe it makes sense to define a `length_getall` that just computes this without actually forming unneeded intermediate `getall` results?

OTHO I find it hard to reason about the optimizer and inference here so not sure if this would actually improve things.

jw3126:
> Ideally, setall should support vectors/tuples/any collection as the values-to-set. This is fundamentally type-stable when the Vector is concretely typed, and otherwise it should work nevertheless.
Do you have some else in mind?

There are some optimizations one can do in the vector case, but that does not have to be in this PR. Like using views for splitting in setall and maybe having a BangBang style `getall!!` implementation that uses `append!` for vectors internally. 

> Haven't tried that, but most likely it has to be yet another wannabe-recursive function generated with macros...
> I'm also thinking that maybe some insights from https://github.com/JuliaObjects/Accessors.jl/pull/23 development could be useful here?

One issue I remember with recursion is this: https://github.com/JuliaLang/julia/issues/43296#issuecomment-991104427
One can workaround it by defining optic style on both the type and the instance:
```julia
OpticStyle(::Type{<:Optic}) = ...
OpticStyle(::Optic) = ...
```
and only call on instances instead of types whenever possible.

jw3126:
> Curious what usecases you foresee for this, and what the semantic should be.

The semantics would be that `getall!!(out, obj, optic)` is equivalent to `append!!(out, getall(obj, optic))`. That is it concatenates `out` with `getall(obj, optic)` by mutation if possible or in a functional way otherwise.

The main use case of `getall!!` is probably internal, namely to implement `getall`. In general, you can use `getall!!`, when you `getall! allocates too much. Also it can be hard for the compiler to infer the output type and passing `out` explicitly can help with that. Typically the user would pass an empty `out`. But non empty out can be useful, for instance to implement `getall` in terms of `getall!!`. 

>  Can it even be defined unambiguously for nontrivial cases? Like, `a |> Elements() |> Elements()`.

Can you give an example what potential ambiguity you see?

bgroenks96:
I am a bit confused as to the relationship between this PR and `getall` to what is implemented in #23. Could someone clarify this?

jw3126:
> I am a bit confused as to the relationship between this PR and `getall` to what is implemented in #23. Could someone clarify this?

The compiler does not like the kind of recursion used in #23. The PRs here infer better but don't allow recursion. For instance you can't get all `Float64` fields burried in an arbitrary nested struct with this PR.

bgroenks96:
But you are able to make it type stable for up to 4 levels of nesting currently? Could this be extended further? There probably is a reasonable limit (maybe <10) that would cover 99% of real world use cases.

jw3126:
I think there is another subtle issue here. What is type stable here are cases, where the recursion depth is "obvious". Like 
`Properties() |> Properties() |> Properities()` is depth 3. But cases like `Recursive(..., Properties())` are hard for inference even if you only use them on depth 3 structs I think.

bgroenks96:
You may want to look at how this problem was solved in Flatten.jl `reconstruct`. I used generated functions to recursively construct a double-linked tree data structure that allowed `reconstruct` to generate code at compile-time that would reconstruct any nested concrete type. It's messy, but it works. And as far as I know, it is still type-stable even on Julia 1.8.

src/getsetall.jl
```diff
@@ -51,26 +96,39 @@ _reduce_concat(xs::AbstractVector) = reduce(append!, xs; init=eltype(eltype(xs))
...
 end
 
+nestedsum(ls::Int) = ls
+nestedsum(ls::Val) = ls
+nestedsum(ls::Tuple) = sum(_val ∘ nestedsum, ls)
+
+to_nested_shape(vs, ::Val{LS}, ::Val{1}) where {LS} = (@assert length(vs) == _val(LS); vs)
 for i in 2:10
-    eval(_generate_getall(i))
+    @eval @generated function to_nested_shape(vs, ls::Val{LS}, ::Val{$i}) where {LS}
```
jw3126, line 118: This is `@eval` and `@generated`?

src/getsetall.jl
```diff
@@ -29,17 +31,60 @@ getall(obj, ::Properties) = getproperties(obj) |> values
 getall(obj, o::If) = o.modify_condition(obj) ? (obj,) : ()
 getall(obj, f) = (f(obj),)
 
+setall(obj, ::Properties, vs::Tuple) = setproperties(obj, NamedTuple{propertynames(obj)}(vs))
+setall(obj::NamedTuple{NS}, ::Elements, vs::Tuple) where {NS} = NamedTuple{NS}(vs)
+setall(obj::NTuple{N, Any}, ::Elements, vs) where {N} = (@assert length(vs) == N; Tuple(vs))
```
jw3126, line 36: Does this actually infer? How about annotating the return type with `::NTuple{N,Any}`?

test/test_getsetall.jl
```diff
@@ -64,6 +64,34 @@ if VERSION >= v"1.6"  # for ComposedFunction
     obj = ([1, 2], [:a, :b])
     @test [1, 2, :a, :b] == @inferred getall(obj, @optic _ |> Elements() |> Elements())
 end
+
+@testset "setall" begin
+    @test (2,) === @inferred setall((1,), Elements(), (2,))
+    @test (2,) === setall((1,), Elements(), [2,])
```
jw3126, line 70: inference is broken here right now?

src/getsetall.jl
jw3126: Ok makes sense thanks! Can you paste your answer as a comment in the code?

src/getsetall.jl
jw3126: Or even better `::NTuple{N,eltype(vs)}`

Project.toml
```diff
@@ -1,7 +1,7 @@
 name = "Accessors"
 uuid = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 authors = ["Takafumi Arakaki <aka.tkf@gmail.com>", "Jan Weidner <jw3126@gmail.com> and contributors"]
-version = "0.1.21"
+version = "0.1.23"
```
jw3126, line 4:
```suggestion
version = "0.1.22"
```

jw3126 (approved):
LGTM thanks a lot @aplavin this is awesome!

PR #69, opened 2022-08-12
more Base lenses

aplavin:


src/setindex.jl
```diff
@@ -5,7 +5,9 @@ end
 @inline setindex(::Base.RefValue, val) = Ref(val)
 
 Base.@propagate_inbounds function setindex(xs::AbstractArray, v, I...)
-    T = promote_type(eltype(xs), typeof(v))
+    T = all(i -> i isa Integer, I) ?
```
jw3126, line 8: Not sure if you are aware, @phipsgabler works on the same issue in Setfield.jl https://github.com/jw3126/Setfield.jl/issues/175 . Your approach looks very slick, but it might miss some corner cases. For instance 
```julia
@set v[CartesianIndex(1)] = :hello
```
will produce a `Vector{Any}` I think?

phipsgabler: Funny, I came up with the exact same solution yesterday, seems like it is a sensible one :D But is your way type stable? I out of habit put the check and selection into a generated function.

phipsgabler: I have proposed two implementations in the issue linked above. But the `I isa Tuple{Vararg{Integer}}` variant seems superior to both.

test/test_core.jl
```diff
@@ -242,6 +242,20 @@ end
...
+    end
+
+    obj = [:a, :b, :c]
+    l = @optic _[1:2]
+    @test @inferred(l(obj))::Vector{Symbol} == [:a, :b]
+    @test @inferred(set(obj, l, [:x, :y]))::Vector{Symbol} == [:x, :y, :c]
+    l = @optic _[CartesianIndex(2)]
+    @test @inferred(l(obj)) == :b
+    @test @inferred(set(obj, l, :x))::Vector{Symbol} == [:a, :x, :c]
+
```
phipsgabler, line 258: Here are some more tests I have used for the slice issue:

```julia
@assert setindex(rand(2,2), zeros(2), 1, :) isa Matrix{Float64}
@assert setindex(Array{Int, 0}(undef), "sdf") isa Array{Any, 0}
@assert setindex([rand(2) for i in 1:2, j in 1:2], zeros(2), 1, 1) isa Matrix{Vector{Float64}}
@assert setindex([rand(2) for i in 1:2, j in 1:2], [:bla], 1, 1) isa Matrix{Vector}
@assert setindex([rand(2) for i in 1:2, j in 1:2], :bla, 1, 1) isa Matrix{Any}
@assert setindex([rand(2) for i in 1:2, j in 1:2], [im, im], 1, :) isa Matrix{Any}
```

phipsgabler: I just dumped them here in case they're useful :D But there are no tests for slicing, as far as I can see, so maybe do include the first and last one?

src/setindex.jl
```diff
@@ -31,3 +32,6 @@ end
 @inline setindex(nt::NamedTuple, v, idx::Tuple{Vararg{Symbol}}) = merge(nt, NamedTuple{idx}(v))
 
 @inline setindex(p::Pair, v, idx::Integer) = Pair(setindex(Tuple(p), v, idx)...)
+
+@inline setindex(x::Number, v, idx::Integer) = (@boundscheck @assert idx == 1; v)
```
jw3126, line 36: I think it is better to throw a `BoundsError`.
```suggestion
@inline function setindex(x::Number, v, idx::Integer)
    @boundscheck if idx != only(eachindex(x))
        throw(BoundsError(x, idx))
    end
    return v
end
```

src/functionlenses.jl
```diff
@@ -53,6 +53,13 @@ set(x,     ::typeof(imag), y) = real(x) + im*y
 set(x,    ::typeof(angle), y) = abs(x) * cis(y)
 set(x,      ::typeof(abs), y) = y >= zero(y) ? y * sign(x) : throw(DomainError(y, "cannot set abs($x) to $y"))
 
+set(x, ::typeof(modf), (fpart, ipart)) = ipart + fpart
+set(x, ::typeof(mod2pi), y) = set(x, Base.Fix2(mod, 2π), y)
+set(x, f::Base.Fix2{typeof(rem)}, y) = f.x * div(x, f.x) + y
+set(x, f::Base.Fix2{typeof(mod)}, y) = f.x * fld(x, f.x) + y
+set(x, f::Base.Fix2{typeof(div)}, y) = y * f.x + rem(x, f.x)
+set(x, f::Base.Fix2{typeof(fld)}, y) = y * f.x + mod(x, f.x)
```
jw3126, line 61: Could also add `mod1` and `cld`? 
Also should we check things like `0<=y < period` for `set(x, @optic(mod(_,period)), y)`?

phipsgabler:
Could you force-push less, please? Doing so makes Github loose track of history and comments...

phipsgabler:
I think the better way to achieve that are rebasing or squashing at the end. But I'm by no means a Git guru, just noticed that Github doesn't properly keep track of comment history and links with force pushes. It's not a big deal though.

jw3126:
I also think there are some tradeoffs with github between nice looking git history and incremental reviewability. Personally, I prefer better reviewability.

jw3126 (approved):
LGTM! Thanks these are nice additions especially fixing slice indexing!

PR #70, opened 2022-08-12
improve invertible lenses

aplavin:
Use the InverseFunctions recommended way to check if a function is invertible: see discussion in https://github.com/JuliaMath/InverseFunctions.jl/issues/13. This check is done just before throwing an exception in `set()`, as a last step if no explicit overloads exist.

Before this PR: hard-coded list of functions to forward to InverseFunctions. Could not cover user- and other package-defined inverses.


jw3126 (approved):
Awesome, less code but more powerful!

PR #77, opened 2022-10-28
add fld, mod, div, rem setters

aplavin:


jw3126 (approved):
Thanks for adding these!

PR #78, opened 2022-12-14
use package extensions for staticarrays

aplavin:
Following https://pkgdocs.julialang.org/dev/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions).
Uses `Requires` as before on 1.8-, and uses the new extensions mechanism instead when it's supported.


jw3126:
Awesome! I never tried the new conditional extensions before.

PR #79, opened 2022-12-21
Add more functionlenses

aplavin:
Some are useful by themselves, some (eg array-related and `convert`) are most convenient when composed with other lenses.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Accessors"
 uuid = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 authors = ["Takafumi Arakaki <aka.tkf@gmail.com>", "Jan Weidner <jw3126@gmail.com> and contributors"]
-version = "0.1.22"
+version = "0.1.23"
 
 [deps]
 Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"


src/functionlenses.jl
```diff
@@ -18,6 +18,11 @@ function set(obj, ::typeof(only), val)
     set(obj, first, val)
 end
 
+function set(x::TX, f::Base.Fix1{typeof(convert)}, v) where {TX}
+    @assert v isa f.x
```
jw3126, line 22: We could probably throw a better error message.

src/functionlenses.jl
```diff
@@ -33,6 +38,37 @@ set(obj::Dict, lens::Union{typeof(keytype),typeof(valtype)}, T::Type) =
...
+################################################################################
+##### ranges
+################################################################################
+set(r::AbstractRange, ::typeof(step), s) = range(first(r), last(r), step=s)
+set(r::AbstractRange, ::typeof(length), l) = range(first(r), last(r), length=l)
+set(r::AbstractRange, ::typeof(first), x) = range(x,  last(r), step=step(r))
+set(r::AbstractRange, ::typeof(last),  x) = range(first(r), x, step=step(r))
+set(r::AbstractUnitRange, ::typeof(first), x) = x:last(r)
+set(r::AbstractUnitRange, ::typeof(last),  x) = first(r):x
+set(r::Base.OneTo, ::typeof(last),  x) = Base.OneTo(x)
```
jw3126, line 70: Behavior is not obvious here for me. For instance one can change the first element of a range in two ways:
* Keep `length` and `step` but change `last`
* Keep `length` and `last` but change `step`

So I am not sure we should have these methods.

test/test_functionlenses.jl
```diff
@@ -124,6 +174,11 @@ end
     o = @optic 1/(1 + exp(-_))
     @test o(2) ≈ 0.8807970779778823
     @test @inferred(set(2, o, 0.999)) ≈ 6.906754778648465
+
+    # setting inverse
+    f = @set inverse(sin) = sqrt
+    @test f(2) == sin(2)
```
jw3126, line 180: I find this super confusing. My expectation would be that `f(x) == x^2` but this seems to test `f(x) = sin(x)`?

jw3126 (commented):
Thanks for adding more and more function lenses! Some of these I find confusing, see my comments above.

test/test_functionlenses.jl
jw3126: Ok I see. To make the test less confusing could you so something like `myasin(x) = asin(x)+2pi` and use that instead of sqrt?

src/functionlenses.jl
jw3126: Yeah these manipulations are useful. If we find a way to disambiguate I would be happy to have them.

jw3126 (approved):
Nice LGTM!

PR #80, opened 2023-01-15
add @accessor macro

aplavin:
Helper macro so that defining simple accessors (get+set ters) doesn't require repetitions, making them less error-prone.

Eg instead of
```julia
my_x(s::MyStruct) = s.x
Accessors.set(s::MyStruct, ::typeof(my_x), v) = @set s.x = v
```
write
```julia
@accessor my_x(s::MyStruct) = s.x
```

Will add docstring if this addition is welcome.

### DIFF
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -398,9 +398,11 @@ macro accessor(ex)
     def = splitdef(ex)
     length(def[:args]) == 1 || throw(ArgumentError("@accessor only supports single argument functions. Overload `Accessors.set(obj, ::typeof($(def[:name])), v)` manually."))
     arg = only(def[:args])
+    argname = splitarg(arg)[1]
+    body_optic = MacroTools.replace(def[:body], argname, :_)
     quote
         $ex
-        $Accessors.set($arg, ::typeof($(def[:name])), v) = $Accessors.@set $(def[:body]) = v
+        $Accessors.set($arg, ::typeof($(def[:name])), v) = $set($argname, $Accessors.@optic($body_optic), v)
     end |> esc
 end
 
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -565,11 +565,13 @@ end
 struct MyStruct
     x
 end
-@accessor my_x(s::MyStruct) = s.x
-@accessor Base.:(+)(s::MyStruct) = -s.x.a
+@accessor my_x(s) = s.x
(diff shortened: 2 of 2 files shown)


jw3126:
Thanks! You are right, there is often code duplication between setters and getters. 
One case I encounter frequently is that I have structs that store equivalent information using
different conventions. Say on struct uses `mm` internally and the other `cm`. In this case I often use the following pattern:
```julia
function optic_length_mm(obj)
    if hasfield(obj, :length_mm)
        @optic _.length_mm
    else
        @optic _.length_cm |> 10 * _
    end
end

length_mm(obj) = optic_length_mm(obj)(obj)
Accessors.set(obj, ::typeof(optic_length_mm), val) = set(obj, optic_length_mm(obj), val)
```
The macro you describe cannot quite cover this, but it could if the wrapped function returns an optic instead of syntax describing an optic. So 
```julia
@accessor my_x(s::MyStruct) = s.x
```
would become
```julia
@accessor my_x(s::MyStruct) = @optic _.x
```
and the above boilerplate would become
```julia
@accessor function length_mm(obj)
    if hasfield(obj, :length_mm)
        @optic _.length_mm
    else
        @optic _.length_cm |> 10 * _
    end
end
```
What do you think?

jw3126:
> Why not
> 
> ```julia
> length_mm(x::StructWithMM) = x.length_mm
> length_mm(x::StructWithCM) = x.length_cm * 10
> ```
> 
> ? Then just add `@accessor` to both lines to make `length_mm` "settable".
> 

In the `if` case I wanted things to be duck typed in order to not depend on `StructWithMM` and `StructWithCM`. But that is probably a special enough situation that the macro does not have to handle.

> `length_mm(obj)` and `set(obj, ::typeof(optic_length_mm), val)` signatures seem less general/composable to me. This approach focuses on cases when `length_mm` has a single method (the one with `if`s) instead of utilizing dispatch.

Dang, I made a typo. It has the same composability if you do it correctly. The correct `set` overload is
```julia
Accessors.set(obj, ::typeof(length_mm), val) = set(obj, optic_length_mm(obj), val)
```
> 
> Also, I'm thinking from the PoV of someone who has working code already, just wants to make it compatible with `Accessors`. It would be great if the only required modification is adding a macro in front of the function definition!

That is a good point.

jw3126 (approved):
LGTM thanks!

PR #85, opened 2023-01-25
More Base functionlenses

aplavin:
Accessors for more functions from Base: first/last two-arg methods, front/tail, string operations (incl `setindex`).

Useful for data manipulation. I gathered these from a couple of my projects, made more consistent, added tests.

### DIFF
--- a/test/test_functionlenses.jl
+++ b/test/test_functionlenses.jl
@@ -186,11 +186,16 @@ end
     @test @inferred(set(2, o, 0.999)) ≈ 6.906754778648465
 
     # parse-related
-    @test @inferred(modify(x -> -2x, "3", @optic parse(Int, _))) == "-6"
-    @test_throws ErrorException modify(log10, "100", @optic parse(Int, _))
-    @test modify(log10, "100", @optic parse(Float64, _)) == "2.0"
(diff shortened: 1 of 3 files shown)


test/test_core.jl
```diff
@@ -309,6 +311,8 @@ end
         @test l(obj) == 20
         @test set(obj, l, true) == (a=(1, (a=10, b=true), 3), b=4)
     end
+
+    @test set("a c", @optic(_[findfirst(' ', _)]), 'b') == "abc"
```
jw3126, line 315: Wow this is really fancy.

test/test_functionlenses.jl
```diff
@@ -156,6 +185,13 @@ end
     @test o(2) ≈ 0.8807970779778823
     @test @inferred(set(2, o, 0.999)) ≈ 6.906754778648465
 
+    # parse-related
+    @test modify(x -> -2x, "3", @optic parse(Int, _)) == "-6"
```
jw3126, line 189: Out of curiosity, does inference handle `@optic parse(Int, _)` well?

jw3126 (commented):
Nice you are really on the roll today!

PR #87, opened 2023-01-26
more package integrations

aplavin:
Add thin layers to support functions/types from AxisKeys, IntervalSets, SkyCoords, StructArrays. This should make `Requires` includes in my `AccessorsExtra` package obsolete on Julia 1.9 - everything will work with just `Accessors` and its package extensions!

### DIFF
--- a/test/test_extensions.jl
+++ b/test/test_extensions.jl
@@ -79,7 +79,8 @@ end
     v = @SVector [1.,2,3]
     @test (@set v[1] = 10) === @SVector [10.,2,3]
     @test (@set v[1] = π) === @SVector [π,2,3]
-    @test (@set v.x = 10) === @SVector [10.,2,3]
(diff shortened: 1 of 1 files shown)


Project.toml
```diff
@@ -16,33 +16,49 @@ StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [weakdeps]
+AxisKeys = "94b1ba4f-4ee9-5380-92f1-94cde586c3c5"
+IntervalSets = "8197267c-284f-5f27-9208-e0e47529a953"
+SkyCoords = "fc659fc5-75a3-5475-a2ea-3da92c065361"
```
jw3126, line 21: The other packages seem to be "general purpose" computing packages, but SkyCoords seems very niche?

PR #89, opened 2023-02-12
fix @accessor on type

aplavin:
Nasty bug: all setters created by @accessor T(x) = ... where T is a type were actually a single set(x, ::DataType, v) method.

Seems like a reasonably clean fix?

### DIFF
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -417,19 +417,23 @@ macro accessor(ex)
     arg = only(def[:args])
     argname = splitarg(arg)[1]
     body_optic = MacroTools.replace(def[:body], argname, :_)
-    ftype = :(
-        if $fname isa Function
-            typeof($fname)
-        elseif $fname isa Type
-            Type{$fname}
-        else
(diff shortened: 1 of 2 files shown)


src/sugar.jl
```diff
@@ -412,13 +412,24 @@ julia> set((a=1, b=2), my_func, 100)
...
     body_optic = MacroTools.replace(def[:body], argname, :_)
+    ftype = :(
+        if $fname isa Function
+            typeof($fname)
+        elseif $fname isa Type
+            Type{$fname}
+        else
+            # is it possible at all?
+            error("Unsupported accessor $(fname)::$(typeof($fname))")
+        end
```
jw3126, line 428:
```suggestion
        if $fname isa Type
            Type{$fname}
        else
            typeof($fname)
        end
```
Not sure if this is better. Would this cover the case, `@accessor (f::MyStruct)(x) = ...`?

jw3126:
yes you can merge this

PR #90, opened 2023-02-24
Add prefix to extension

devmotion:
Extensions with non-unique name cause collisions and hence break e.g. compilation of sysimages. This PR adds a prefix to the extension to work around that issue. See https://github.com/JuliaMath/ChangesOfVariables.jl/pull/13 for a longer discussion.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -19,7 +19,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 
 [extensions]
-StaticArraysExt = "StaticArrays"
+AccessorsStaticArraysExt = "StaticArrays"
 
 [compat]
 Compat = "3.18, 4"
--- a/ext/AccessorsStaticArraysExt.jl
+++ b/ext/AccessorsStaticArraysExt.jl
@@ -1,4 +1,4 @@
-module StaticArraysExt
+module AccessorsStaticArraysExt
 isdefined(Base, :get_extension) ? (import StaticArrays) : (import ..StaticArrays)
 using Accessors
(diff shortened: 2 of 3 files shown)


jw3126:
Thanks a lot! Can you also prefix the other extensions?

devmotion:
Well, I assume this is desired even if the name collision issues are fixed, see e.g. https://github.com/JuliaMath/ChangesOfVariables.jl/pull/13#issuecomment-1427763012.

devmotion:
>  Some official scheme of extension naming would be nice.

I guess it's already established inofficially, most packages that I'm aware of use these prefixes (e.g., SciML, SpecialFunctions, and LogExpFunctions).

devmotion:
Github doesn't show any results, is the link you posted correct? In any case, I'm sure there are packages which don't use prefixes but a major part of the ecosystem, including core packages, already does. And I guess it's rather these packages which form inofficial standards.

devmotion:
> Just clicked again - seems to work. It's search for `[weakdeps]` across all TOML-type files.

That's what I assumed but it doesn't work for me:
image

>  And this prefixing leads to redundancy in messages, that's not helpful.

I wonder if that might be (partly) a complaint about the standard Pkg messages rather than the names of the extensions. I imagine that if there is an error in one of the extension modules then standard logging/error messages might only point you to `ChainRulesCoreExt`, and it might not be immediately clear to which parent package the extension belongs.

Generally, I'm not surprised that quite a few packages don't use prefixes yet. Many packages switched to prefixes only very recently (e.g., SpecialFunctions a few days ago), now that it became apparent that there are benefits.

jw3126 (approved):
LGTM. We should add the `Accessors` prefix for now, if it is not needed in future anymore, we can always drop it again. Thanks @devmotion

PR #93, opened 2023-03-28
Show essentials of package in readme

KronosTheLate:
This is because I was really expecting `@set` to overwrite the old variable, and so I have documented the part that confused me in the readme in a simple example.

I know this is spelled out in the docs, but I honestly expect to be able to use such a simple package by seing 2 lines of example code, and do not come to it with the mindset that it is neccessary to open docs. In case others are like me, this should make their lives easier.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,7 +30,7 @@ julia> nt
 (a = 1, b = 2)
 ```
 
-To overwrite the old definition, we can rebind `nt` to the mutated version:
+To overwrite the old definition, we can rebind `nt` to the new version:
 ```julia
 julia> nt = @set nt.b=3
 (a = 1, b = 3)
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ julia> @set nt.b=3
 (a = 1, b = 3)
 ```
 
-Note that this only returns a updated copy of `nt`, and *does not overwrite or mutate* the 
+Note that this only returns an updated copy of `nt`, and *does not overwrite or mutate* the 
 value bound to `nt`:
 ```julia
 julia> nt
--- a/README.md
+++ b/README.md
@@ -23,7 +23,7 @@ julia> @set nt.b=3
 (a = 1, b = 3)
 ```
 
-Note that this only returns a mutated version of `nt`, and *does not overwrite* the 
+Note that this only returns a updated copy of `nt`, and *does not overwrite or mutate* the 
 value bound to `nt`:
 ```julia
 julia> nt
--- a/README.md
+++ b/README.md
@@ -8,12 +8,47 @@ The goal of [Accessors.jl](https://github.com/JuliaObjects/Accessors.jl) is to m
 It is the successor of [Setfield.jl](https://github.com/jw3126/Setfield.jl).
 
 # Usage
-Updating immutable data was never easier:
+Say you have some immutable data structure, such as a `NamedTuple`:
 ```julia
-using Accessors
-@set obj.a.b.c = d
+julia> nt = (a=1, b=2)
+(a = 1, b = 2)
+
 ```
-To get started, see [this tutorial](https://juliaobjects.github.io/Accessors.jl/stable/getting_started/) and/or watch this video:
+If you try something like `nt.b=3`, it will throw an error. But 
+using Accessors, we can change it anyways:
+```julia
+julia> using Accessors
+
+julia> @set nt.b=3
+(a = 1, b = 3)
+```
+
+Note that this only returns a mutated version of `nt`, and *does not overwrite* the 
+value bound to `nt`:
+```julia
+julia> nt
+(a = 1, b = 2)
+```
+
+To overwrite the old definition, we can rebind `nt` to the mutated version:
+```julia
+julia> nt = @set nt.b=3
+(a = 1, b = 3)
+
+julia> nt
+(a = 1, b = 3)
+```
+
+As this is a common use case, the convenience macro `@reset` overwrites what you are mutating:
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -8,12 +8,47 @@ The goal of [Accessors.jl](https://github.com/JuliaObjects/Accessors.jl) is to m
 It is the successor of [Setfield.jl](https://github.com/jw3126/Setfield.jl).
 
 # Usage
-Updating immutable data was never easier:
+Say you have some immutable data structure, such as a `NamedTuple`:
 ```julia
-using Accessors
-@set obj.a.b.c = d
```
jw3126, line 14: Can you keep the original three lines as TLDR?

jw3126 (approved):
Thanks a lot, this is a nice improvement.

README.md
KronosTheLate: That is possible, but I am hesitant about it. It still looks to me like it should update `obj`, so if I only read that, I find it misleading. But having the mutating version in a line below should alert the user to check out both, something like 
```
using Accessors
@set obj.a.b.c = d     # Non-mutating
@reset obj.a.b.c = d   # Mutating
```
would resolve that issue for me. Do you find that to be a good solution? I still feel like that splits the "tutorial" in 3 - the TLDR, the longer version below, and the actual docs. It feels like too much fracturing in documentation for a quite simple package to me... But if you much prefer it, it is fine by me.

Slightly separate but related issue; the name `reset` does not sit well with me. I intuitively expect it to change the fields changed by `@set` back to their initial value, resetting them. The current `@reset` does not in fact "reset" anything. I would much prefer `@set` and `@set!`. Is such an API change feasible? Or at least alias `@set!` to reset? (Or the other way around).

KronosTheLate: If we agree that `@set!` is better than `@reset`, I suggest the following:
Define `@set!` to do what `@reset` currently does (non-breaking), bind `const @reset = @set!`, print deprecation warnings when `@reset` is used (because in the long term, we do not want 2 macros doing the same thing) for some months/a year, and then deprecate `@set!` in some future breaking release.

jw3126: >  I still feel like that splits the "tutorial" in 3 - the TLDR, the longer version below, and the actual docs.

Good point, we can drop the TLDR;

> If we agree that `@set!` is better than `@reset`

In Setfield.jl we have `@set!` instead of `@restet` and I consider it a bad name. To me it is important to distinguish between rebinding a variable and actual mutation. Mutation is a notorious source of complexity and bugs, while rebinding a variable is a very local change and does not cause problems in my experience.

README.md
```diff
@@ -8,12 +8,47 @@ The goal of [Accessors.jl](https://github.com/JuliaObjects/Accessors.jl) is to m
...
+If you try something like `nt.b=3`, it will throw an error. But 
+using Accessors, we can change it anyways:
+```julia
+julia> using Accessors
+
+julia> @set nt.b=3
+(a = 1, b = 3)
+```
+
+Note that this only returns a mutated version of `nt`, and *does not overwrite* the
```
jw3126, line 26:
```suggestion
Note that this only returns a updated copy of `nt`, and *does not overwrite or mutate* the 
```

README.md
```diff
@@ -8,12 +8,47 @@ The goal of [Accessors.jl](https://github.com/JuliaObjects/Accessors.jl) is to m
...
+```
+
+Note that this only returns a mutated version of `nt`, and *does not overwrite* the 
+value bound to `nt`:
+```julia
+julia> nt
+(a = 1, b = 2)
+```
+
+To overwrite the old definition, we can rebind `nt` to the mutated version:
```
jw3126, line 33:
```suggestion
To overwrite the old definition, we can rebind `nt` to the new version:
```

README.md
```diff
@@ -8,12 +8,47 @@ The goal of [Accessors.jl](https://github.com/JuliaObjects/Accessors.jl) is to m
...
+```julia
+julia> nt = @set nt.b=3
+(a = 1, b = 3)
+
+julia> nt
+(a = 1, b = 3)
+```
+
+As this is a common use case, the convenience macro `@reset` overwrites what you are mutating:
```
jw3126, line 42:
```suggestion
As this is a common use case, the convenience macro `@reset` rebinds the updated value:
```

README.md
KronosTheLate: I see! Alright, so `@set!` is misleading, as nothing is actually mutated.

However, we still have a problem to fix IMO: `@reset` makes it sound like something is going back to some original state, i.e. being "reset". This is misleading, as the new state is previously unseen. How about the quite plain and accurate `@rebind`, with as many chars as `@reset`?

README.md
KronosTheLate: perhaps "rebinds the variable to the updated value"? is a bit more explicit?

README.md
jw3126: I agree that `@reset` is not ideal. `@rebind` might be a bit better, but I don't find it to be better by a large enough margin to make a change.

jw3126: `@selfset` was another candidate name I considered but did not find good enough to justify the change.

jw3126:
LGTM, should I merge?

README.md
KronosTheLate: Reset is also similar enough to set that it add some consistency, so it is fint by me

KronosTheLate:
> LGTM, should I merge?

All green lights from me, thanks for the attentive and open attitude to my suggestions ^_^

PR #94, opened 2023-03-31
optics broadcast as scalars, same as regular functions

aplavin:
Optics are analogous to functions, and should broadcast the same way - as scalars.

### DIFF
--- a/test/test_optics.jl
+++ b/test/test_optics.jl
@@ -96,4 +96,10 @@ end
     @test PropertyLens(:a) === PropertyLens{:a}()
 end
 
+@testset "broadcasting" begin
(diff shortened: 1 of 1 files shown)


src/optics.jl
```diff
@@ -465,6 +465,12 @@ end
     delete(obj, IndexLens(lens.f(obj)))
 end
 
+
+Broadcast.broadcastable(
+    o::Union{PropertyLens,IndexLens,DynamicIndexLens,Elements,Properties,If,Recursive}
+) = Ref(o)
```
jw3126, line 471:
```suggestion
) = Ref(o)
```
I wonder if `Ref` might allocate sometimes? Maybe `(o,)` is better?

jw3126 (approved):
LGTM, thanks!

PR #95, opened 2023-03-31
getall, setall: fix and cleanup

aplavin:
- Bugfix: only the default composition order was supported in `getall`/`setall`
- Cleanup, throw a more intuitive error


jw3126 (approved):
LGTM, thanks for fixing this!

PR #97, opened 2023-03-31
Indexing

aplavin:
Improve support of index-based optics:
- More types supported in `set` and `delete` for indices (& first/last)
- Generalize parsing of dynamic indexing, so that `@optic(_[[end-2, end]])` works


jw3126 (approved):
LGTM, thanks!

Issue #14, opened 2020-12-11
`Leaves` optic


cscherrer:
@jw3126 I think this is working well now. What do you think of moving `Leaves` into Accessors? It's very fast, and seems it could be useful for others.

Most of the code is [here](https://github.com/cscherrer/NestedTuples.jl/blob/master/src/leaves.jl). Biggest concern I could see is that it depends on GeneralizedGenerated. There may be a way to drop that dependency, but it would take some fiddling.

It's no trouble to me either way, and I won't be offended if you don't think it's a good fit. Just seems more natural to me to have it in Accessors.

jw3126:
Nice! Accessing the leaves of a tree is certainly an important task. As I understand, this assumes that the tree structure is encoded using `Tuples` and `NamedTuples`. There are many other possible choices, however. What if I have a tree like this:
```julia
["some", ["tree", [["encoded", "using"], "vectors"]
```
Is there a way to let the user supply the tree structure?

cscherrer:
Hmm, good question. It's not clear to me in this case even what we'd want the result to look like. Julia infers the type as `Array{Any, 1}`, so that doesn't help.

OTOH, if we instead had some nested structs, that should be manageable. StructArrays has some nice tricks for recasting a struct as a named tuple, and after that it reduces to the previous case.

My interest here was to just work with the part that can be manipulated very quickly. So one approach could be to use this as an "outer optic" and pass to something else for each resolved component.

Then I guess part of the question is whether the current results should be considered "leaves". Maybe `TupleLeaves` or something is a better name?

jw3126:
> Hmm, good question. It's not clear to me in this case even what we'd want the result to look like. Julia infers the type as `Array{Any, 1}`, so that doesn't help.

I would expect bad performance in this case, but the semantics would be:
```julia
tree = ["some", ["tree", [["encoded", "using"], "vectors"]
modify(uppercase, tree, Leaves(...)) == ["SOME", ["TREE", [["ENCODED", "USING"], "VECTORS"]
```
My point was that there are many ways trees are encoded and instead of only supporting one somewhat arbitrary way (descent into `Tuples` and `NamedTuples`) I would prefer if Accessors.jl could be helpful for all these cases.

> My interest here was to just work with the part that can be manipulated very quickly. So one approach could be to use this as an "outer optic" and pass to something else for each resolved component.

[...]

PR #98, opened 2023-04-02
getall: cleaner, support Recursive

aplavin:
Here, I try out a different `getall` implementation based on the trick from https://github.com/FluxML/Functors.jl/pull/61.
It:
- doesn't require code generation (one internal `@generated` function remains, but maybe it can be removed?),
- support `Recursive`, even easier than `ComposedFunction`!
Can potentially close the `getall` part of https://github.com/JuliaObjects/Accessors.jl/issues/63, https://github.com/JuliaObjects/Accessors.jl/issues/14, https://github.com/JuliaObjects/Accessors.jl/pull/23.

The `getall` limit for `ComposedFunction` inference seems to become lower than before, though. It affects one test that tested the maximal depth of 10.
Well, the `getall(ComposedFunction)` implementation may stay as before with `@eval`, only adding `Recursive` from this PR...

@jw3126 @rafaqz @cscherrer

### DIFF
--- a/src/getsetall.jl
+++ b/src/getsetall.jl
@@ -101,6 +101,18 @@ end
     f.(t, Ref(val))
 end
 
+function getall(obj, or::Recursive)
+    recurse(o) = _walk_getall_rec(var"#self#", o, or)
+    _walk_getall_rec(recurse, obj, or)
+end
+_walk_getall_rec(recurse, obj, or::Recursive) =
+    if or.descent_condition(obj)
+        _getall_rec(recurse, obj, or.optic)
+    else
+        (obj,)
+    end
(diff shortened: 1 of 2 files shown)


bgroenks96:
Hey @aplavin what happened with this PR? Is there something blocking it?

aplavin:
For reducing code generation (`@eval`) in `getall`: there's no pressing need to do that, and apparently the compiler still cannot optimize it in all cases:

> The getall limit for ComposedFunction inference seems to become lower than before, though. It affects one test that tested the maximal depth of 10.

---

For adding `Recursive` support: it can be done I guess, but is it needed? Anyone using `Recursive` at all?
IME, it's quite hard to use in practice; I can recommend `RecursiveOfType` from AccessorsExtra.jl instead.

PR #101, opened 2023-04-20
fix setall(Tuple, Properties)

aplavin:
Didn't work: cannot construct a NamedTuple out of Tuple properties, they don't have names.


jw3126 (approved):
LGTM thanks!

PR #102, opened 2023-04-20
fix optics show and its tests

aplavin:
- show on composedoptics didn't work on Julia 1.5- at all
- now run basic show tests on all Julia versions
- add minor broken test as a reminder

### DIFF
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -430,26 +430,26 @@ macro accessor(ex)
     end |> esc
 end
 
-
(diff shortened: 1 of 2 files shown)


jw3126 (approved):
Thanks! Out of curiosity, do you personally use old julia versions where composed function does not show nicely?

jw3126:
Oh I see. BTW I am also happy to drop anything pre 1.6 if it causes trouble.

PR #103, opened 2023-04-20
"flipped" indexlens

aplavin:
Works with dictionaries as well, not only arrays. Pretty convenient:
```julia
julia> lut = dictionary([:a => 1, :b => 2, :c => 3])
julia> obj = (x=:a, y=2)
julia> o = @optic lut[_.x]

julia> o(obj)
1
julia> modify(x -> x + 1, obj, o)
(x = :b, y = 2)
```

Hope I don't overwhelm you with multiple PRs :) Each is pretty self-contained and independent, so a couple of smaller ones should be better than a single large.

### DIFF
--- a/src/functionlenses.jl
+++ b/src/functionlenses.jl
@@ -42,12 +42,6 @@ function set(obj::NamedTuple, ::Type{NamedTuple{KS}}, val::NamedTuple) where {KS
     setproperties(obj, NamedTuple{KS}(val))
 end
 
-function set(obj, f::Base.Fix1{typeof(getindex)}, val)
-    ix = findfirst(isequal(val), f.x)
-    ix === nothing && throw(ArgumentError("value $val not found in $(f.x)"))
-    return ix
-end
-
 ################################################################################
 ##### eltype
 ################################################################################
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -319,12 +319,9 @@ end
 end
 
 @testset "flipped index" begin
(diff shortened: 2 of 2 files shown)


src/functionlenses.jl
```diff
@@ -42,6 +42,12 @@ function set(obj::NamedTuple, ::Type{NamedTuple{KS}}, val::NamedTuple) where {KS
     setproperties(obj, NamedTuple{KS}(val))
 end
 
+function set(obj, f::Base.Fix1{typeof(getindex)}, val)
+    ix = findfirst(isequal(val), f.x)
+    ix === nothing && throw(ArgumentError("value $val not found in $(f.x)"))
+    return ix
+end
```
jw3126, line 49: I have some objections. 
* `findfirst` is typically `O(n)`, preferably optics are fast 
* If there are multiple matches, I would prefer an error.

jw3126:
How about doing this instead:
```julia
struct Bijection{T}
    table::T
end
(f::Bijection)(key) = table[key]
function Accessors.set(old_key, f::Bijection, new_val) 
    # findfirst ....
end

f = Bijection(lut)
o = @optic f(_.x)
```

jw3126:
With this suggestion you wouldn't use square brackets so no new parsing needed I think?

jw3126:
I am against adding this parsing rule. It adds complexity IMO without enough gains.
Turning a dict into a function seems a natural operation. I could have sworn I once used a package for this but don't remember which.

PR #108, opened 2023-05-15
improve getall/setall for staticarrays

aplavin:
Now, it's more consistent:
- getall comptime-known length: Tuple
- getall runtime length: Vector

This change lets them nicely propagate and infer for both get- and setall.

Non-breaking: container types were specifically allowed to change in getall docsting.

### DIFF
--- a/test/test_getsetall.jl
+++ b/test/test_getsetall.jl
@@ -41,16 +41,16 @@ if VERSION >= v"1.6"  # for ComposedFunction
     @test (2, 5, 10, 17, 26, 37) === @inferred getall(obj, @optic _ |> _[:] |> Elements() |> Elements() |> _[:] |> Elements() |> Elements() |> _[1]^2 + 1 |> only)
 
     # trickier types for Elements():
-    obj = (a=("ab", "c"), b=([1 2; 3 4],), c=(SVector(1), SVector(2, 3)))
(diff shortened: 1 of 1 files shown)


Project.toml
```diff
@@ -57,4 +59,5 @@ Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
 Aqua = "4c88cf16-eb10-579e-8560-4a9242c79595"
 
 [targets]
-test = ["Test", "PerformanceTestTools", "QuickTypes", "InverseFunctions", "AxisKeys", "IntervalSets", "StaticArrays", "StructArrays", "BenchmarkTools", "InteractiveUtils", "StaticNumbers", "Unitful","Aqua"]
+test = ["Test", "PerformanceTestTools", "QuickTypes", "InverseFunctions", "AxisKeys", "IntervalSets", "StaticArrays", "StructArrays", "BenchmarkTools", "InteractiveUtils", "StaticNumbers", "Unitful", "Aqua", "ConstructionBaseExtras"]
```
jw3126, line 62: Does not have to be this PR, but `test` gets quite crowded. Maybe we should have `test/Project.toml`.

jw3126: Or maybe it is fine. It won't shrink the `extra` section much, since many of these are extensions anyway?

jw3126 (approved):
LGTM

PR #109, opened 2023-05-15
remove StaticArrays from hard deps

aplavin:
It's really not needed for anything, neither on 1.9- nor on 1.9.
I added it in https://github.com/JuliaObjects/Accessors.jl/pull/78 just by mistake, and after that StaticArrays were always installed with Accessors on 1.9-.


jw3126:
I think we can merge this.

PR #121, opened 2024-01-04
use weakdeps wherever possible

aplavin:
No behavior change on any Julia version, just transform some deps into weakdeps on 1.9+.


jw3126:
Mhh what is the benefit of making stdlibs into extensions?
I wonder if this creates friction when you do something like `cd Accessors; ]activate .; include("test/runtests.jl")`

jw3126:
> Well, this doesn't work as-is anyway: it requires test dependencies from `test/Project.toml`, not just regular dependencies. So no extra friction: proper test env activation is needed anyway.

Ok that is a good argument :smile:

jw3126:
Ok thanks, we can merge this!

PR #122, opened 2024-01-04
fix tests on 1.10

aplavin:
Only one minor regression: in allocations when the compose order is right-associative


jw3126:
Thanks a lot!

PR #125, opened 2024-01-04
add shortcuts for Elements(), Properties(), @optic

aplavin:
Allows for shorter and much nicer to read:
- `@optic _[∗].a` instead of `@optic _ |> Elements() |> _.a`,
- `@optic _[∗ₚ][2]` instead of `@optic _ |> Properties() |> _[2]`.

### DIFF
--- a/test/test_optics.jl
+++ b/test/test_optics.jl
@@ -106,6 +106,12 @@ end
     @test (@o _.a[2]) === (@optic _.a[2])
     @test (@optic _[∗]) === Elements()
     @test (@optic _.a[∗][2]) === (@optic _.a |> Elements() |> _[2])
+    @test (@optic _.a[∗ₚ][2]) === (@optic _.a |> Properties() |> _[2])
+    # user-defined symbols have priority, same as elsewhere in Julia
+    let ∗ = 3
+        o = @optic _[∗]
+        @test o([1,2,42]) == 42
+    end
 end
 
 end#module


jw3126:
I wished several times, that these were easier to type. My main issue is that they require so many pipes and underscores around them. I wonder if we could do something like `@optic _.@Elements().myfield`. Possibly with a shorter name than `@Elements`.  Or another mechanism that `@` that makes chaining easier to type. 
I also wonder if we should have `@o` as a shorthand for `@optic`, I think I saw that already in one of your packages.

One downside of the current proposal is that it collides with `getindex`.

jw3126:
btw thanks a lot for all the work you put into accessors already this year :smile:

jw3126:
> Especially for `Elements`, `getindex` seems quite natural visually: `xxx[∗]`. What exactly is the "collision" you have in mind? Note that `∗` (`\ast`) is not `*`, so it's not type piracy! It's a unicode shortcut, I think it's fine to have unicode symbols as shortcuts as long as it's not the only way.

```julia
∗= 3
f = @optic _[∗]
@assert f([1,2,42]) == 42
```
is a gotcha.

jw3126:
> Your example is similar to someone defining a variable/function/... named `Elements` or `If`: the effect is the same, corresponding `Accessors` exports become shadowed in this case. That's just how Julia works.

In these cases the correct scope is used:
```julia
using Accessors

struct Elements end

(::Elements)(x) = 2*x


o = @optic _ |> Elements()

@assert o(1)  == 2 # works
```

src/sugar.jl
```diff
@@ -430,11 +430,24 @@ macro accessor(ex)
...
 
+### shortcuts:
+# optic macro
+const var"@o" = var"@optic"
+
+# Elements, Properties
+const ∗ = Elements()
+const ∗ₚ = Properties()
+IndexLens(::Tuple{Elements}) = Elements()
+IndexLens(::Tuple{Properties}) = Properties()
```
jw3126, line 441: This looks much better than doing it at "parse time". Can you add a test where the user defines these symbols?

jw3126:
Oh I think I misread your code from the very begining. The problem I imagined with symbols being pulled from the wrong scope might never have existed. Can you add a testcase like this:
```julia
∗= 3
f = @optic _[∗]
@test f([1,2,42]) == 42
```
and then it is good to go.

src/sugar.jl
jw3126: Optionally we could add a layer of indirection and instead of overloading the `IndexLens` constructor with something that is not an IndexLens, do this dispatch in another function `index_like_lens`.

jw3126:
I also like it.

PR #132, opened 2024-02-07
Support `insert` with dynamic `DynamicIndexLens`

MasonProtter:
Before:
```julia
julia> let v = [1,2,3]
           @insert v[end+1] = 4
       end
ERROR: MethodError: no method matching insert(::Vector{Int64}, ::Accessors.DynamicIndexLens{var"#6#7"}, ::Int64)
```
after:
```julia
julia> let v = [1,2,3]
           @insert v[end+1] = 4
       end
4-element Vector{Int64}:
 1
 2
 3
 4
```

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Accessors"
 uuid = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 authors = ["Takafumi Arakaki <aka.tkf@gmail.com>", "Jan Weidner <jw3126@gmail.com> and contributors"]
-version = "0.1.35"
+version = "0.1.36"
 
 [deps]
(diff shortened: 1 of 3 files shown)


jw3126 (approved):
LGTM thanks a lot @MasonProtter. Actually surprised that this did not came up before.

MasonProtter:
Happy to help!

MasonProtter:
What is nice about that property?

MasonProtter:
I guess I'd say this sort of use-case this seems to be the entire point of `DynamicIndexLens`, but I'm nowhere near as familiar with the codebase or what could go wrong if that law isn't followed.

MasonProtter:
@jw3126 any thoughts on @aplavin's point?

jw3126:
Yeah I think promoting `last` is the better option here.

MasonProtter:
Sounds good.

PR #136, opened 2024-02-26
laws testing: add insert/delete, generalize modify

aplavin:


jw3126 (approved):
LGTM

jw3126 (commented):
I also wonder if we should include the `test_modify_law(identity, ...)` in `test_getset_laws` and `test_getsetall_laws`.

PR #138, opened 2024-02-26
improve inference/code generation

aplavin:
Hard to make a minimal reproducer for this, but the setall change makes autodiff (Enzyme) work in some cases it didn't before.

Draft for now, will possibly add more similar changes soon.


jw3126 (approved):
Feel free to merge if it helps your usecase. A test would be nice, but I understand that it is to hard to create an MWE. I have no intuition whether the new or old implementation is better. But maybe add a comment that links to this PR.

jw3126:
If you had a reproducer, might also be worth opening an issue in julia.

PR #139, opened 2024-02-26
show optics as @o

aplavin:


jw3126:
This is a good idea :+1:

jw3126:
We should also mention `@o` in the docstring.

jw3126 (approved):
LGTM, thanks!

PR #155, opened 2024-07-02
nicer show for optics

aplavin:
Optics already show quite nicely in many cases, with a major exception of operators. Here, I make operators display nicer, while staying conservative to ensure robust show-parse roundtrip:
```julia
julia> @o _.a + 1
(@o +(_.a, 1)) # before
(@o _.a + 1) # after

julia> @o _.a*2 + 1
(@o +(*(_.a, 2), 1)) # before
(@o (_.a * 2) + 1) # after
```
Most useful whenever optics are a part of the user-facing interface (can even be plot labels!), but for internal debug/inspection as well.

### DIFF
--- a/test/test_core.jl
+++ b/test/test_core.jl
@@ -462,6 +462,8 @@ end
 @testset "full and compact show" begin
     @test sprint(show, (@optic _.a)) == "(@o _.a)"
     @test sprint(show, (@optic _.a + 1)) == "(@o _.a + 1)"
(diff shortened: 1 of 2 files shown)


test/test_core.jl
```diff
@@ -461,6 +461,7 @@ end
 
 @testset "full and compact show" begin
     @test sprint(show, (@optic _.a)) == "(@o _.a)"
+    @test sprint(show, (@optic _.a + 1)) == "(@o _.a + 1)"
```
jw3126, line 464: Can we add a test where operator precedence plays a role?

PR #159, opened 2024-07-26
Move custom error hint for insert to a package extension

MasonProtter:
This is a followup PR to https://github.com/JuliaObjects/Accessors.jl/pull/140

On modern versions of julia, moving this code to a package extension should have no observable changes, but it makes it so that Accessors.jl doesn't unconditionally load Markdown.jl, and only ends up using it *if* it's already loaded (which will be the case for interactive sessions of julia, but might *not* be the case if one is doing something like making a small binary and are stripping stdlibs. 

I'd argue that in such a situation, this error hint isn't necessary or desirable anyways.


jw3126:
I agree with @aplavin lets keep it simple stupid.

PR #161, opened 2024-07-27
fix  Cannot parse optic _ && true #160 

### LINKED ISSUE #160: Cannot parse optic `_ && true`
```jl
julia> @o (_ && true)
ERROR: LoadError: ArgumentError: Cannot parse optic _ && true. Lens expressions must start with _, got $(Expr(:escape, :(_ && true))) instead.
```

What's happening here?

jw3126:
fix #160


jw3126:
> It doesn't affect optics that happen to include the actual symbol, right?
> 
> ```julia
> julia> :&&
> :&&
> ```

You mean like this?
```
julia> @o (_[:&&])
(@o _[:&&])
```

PR #162, opened 2024-07-30
fix zeros in hypot and norm

aplavin:


ext/AccessorsLinearAlgebraExt.jl
```diff
@@ -4,7 +4,10 @@ import Accessors: set, @set
 using LinearAlgebra: norm, normalize, diag, diagind
 
 set(arr, ::typeof(normalize), val) = norm(arr) * val
-set(arr, ::typeof(norm), val)      = map(Base.Fix2(*, val / norm(arr)), arr) # should we check val is positive?
+function set(arr, ::typeof(norm), val)
+    omul = iszero(val) ? one(norm(arr)) : norm(arr)
+    map(Base.Fix2(*, val / omul), arr)
+end
```
jw3126, line 10: `oneunit` instead? Or should we error? This seems to break lens laws.

jw3126: If you set norm to zero and try to set it back to a non zero value it breaks lens laws. But maybe that is okay.

jw3126: Yeah it is fine sry

Issue #127, opened 2024-01-09
Errors in extensions during precompilation 


ronisbr:
Hi!

Recently, I started to see errors during precompilation of my packages that use Accessors:

```julia
Precompiling SatelliteToolbox
  21 dependencies successfully precompiled in 12 seconds. 35 already precompiled.
  3 dependencies had output during precompilation:
┌ Accessors → AccessorsTestExt
│  ┌ Warning: Module AccessorsDatesExt with build ID ffffffff-ffff-ffff-0000-0128c163477b is missing from the cache.
│  │ This may mean AccessorsDatesExt [308068fd-b978-57fd-b78f-2a52fb63dba4] does not support precompilation but is imported 
by a module that does.
│  └ @ Base loading.jl:1942
│  ┌ Error: Error during loading of extension AccessorsDatesExt of Accessors, use `Base.retry_load_extensions()` to retry.
│  │   exception =
│  │    1-element ExceptionStack:
│  │    Declaring __precompile__(false) is not allowed in files that are being precompiled.
│  │    Stacktrace:
│  │      [1] _require(pkg::Base.PkgId, env::Nothing)
│  │        @ Base ./loading.jl:1946
│  │      [2] __require_prelocked(uuidkey::Base.PkgId, env::Nothing)
│  │        @ Base ./loading.jl:1806
│  │      [3] #invoke_in_world#3
│  │        @ Base ./essentials.jl:921 [inlined]
│  │      [4] invoke_in_world
│  │        @ Base ./essentials.jl:918 [inlined]
│  │      [5] _require_prelocked
│  │        @ Base ./loading.jl:1797 [inlined]
│  │      [6] _require_prelocked
│  │        @ Base ./loading.jl:1796 [inlined]
│  │      [7] run_extension_callbacks(extid::Base.ExtensionId)
│  │        @ Base ./loading.jl:1289
│  │      [8] run_extension_callbacks(pkgid::Base.PkgId)
│  │        @ Base ./loading.jl:1324
│  │      [9] run_package_callbacks(modkey::Base.PkgId)
│  │        @ Base ./loading.jl:1158
│  │     [10] __require_prelocked(uuidkey::Base.PkgId, env::String)
│  │        @ Base ./loading.jl:1813
│  │     [11] #invoke_in_world#3
│  │        @ Base ./essentials.jl:921 [inlined]
│  │     [12] invoke_in_world
│  │        @ Base ./essentials.jl:918 [inlined]
[...]

PR #164, opened 2024-08-10
Don't make Test a dependency

ChrisRackauckas:
It should only be a test dependency

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -12,7 +12,6 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
-Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [weakdeps]
 AxisKeys = "94b1ba4f-4ee9-5380-92f1-94cde586c3c5"
@@ -49,3 +48,4 @@ IntervalSets = "8197267c-284f-5f27-9208-e0e47529a953"
 StaticArrays = "90137ffa-7385-5640-81b9-e52037218182"
 StructArrays = "09ab397b-f2b6-538f-b94a-2f83cf4a842a"
 Unitful = "1986cc42-f94f-5a68-af5c-568840ba703d"
+Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"


longemen3000:
Tests fails because `Test` is loaded here:
https://github.com/JuliaObjects/Accessors.jl/blob/6087e2f8b871708d3c357bd71ceac6c5b0126764/src/Accessors.jl#L22

ChrisRackauckas:
I set the repo up for the new v1.10 LTS

Project.toml
```diff
@@ -12,7 +12,6 @@ LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 Requires = "ae029012-a4dd-5104-9daa-d747884805df"
```
ChrisRackauckas, line 14:
```suggestion
```

ChrisRackauckas:
Looks like that failure is on master and unrelated to this change.

jw3126:
Thanks @ChrisRackauckas We had some issues with extensions before, see 
https://github.com/JuliaObjects/Accessors.jl/issues/127
and 
https://github.com/JuliaLang/julia/issues/52132

If/once https://github.com/JuliaLang/julia/issues/52132 is backported to the LTS we can move stuff to extensions.

devmotion:
On the contrary, the problem with circular loading of extensions only occurs if a one of the transitive dependencies triggers an extension. So if the ecosystem moves to making Test a weak dependency on Julia >= 1.9 but Accessors keeps it as a dependency, then an indirect dependency on Accessors can provoke the circular loading (which apparently was reported before in #127). See e.g. https://github.com/JuliaMath/InverseFunctions.jl/issues/51#issuecomment-2298558303 and the linked issue.

Project.toml
```diff
@@ -37,11 +35,10 @@ IntervalSets = "0.7"
 InverseFunctions = "0.1.5"
 MacroTools = "0.5"
 Markdown = "1"
-Requires = "0.5, 1.0"
 StaticArrays = "1"
 StructArrays = "0.6"
 Unitful = "1"
-julia = "1.6"
+julia = "1.10"
```
devmotion, line 41: IMO this should be
```suggestion
julia = "1.9"
```
so also people on Julia 1.9 can benefit from this fix in case the Test dependency caused circular loading for them.

Project.toml
```diff
@@ -11,8 +11,6 @@ InverseFunctions = "3587e190-3f89-42d0-90ee-14403ec27112"
 LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
 MacroTools = "1914dd2f-81c6-5fcd-8719-6d5c9610ff09"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
-Requires = "ae029012-a4dd-5104-9daa-d747884805df"
-Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
```
devmotion, line 15: Test has to be added to `[weakdeps]`, `[compat]`, and `[extras]`, and a corresponding section has to be added to `[extensions]`

src/Accessors.jl
```diff
@@ -4,50 +4,35 @@ using MacroTools: isstructdef, splitstructdef, postwalk
...
 include("setindex.jl")
 include("optics.jl")
 include("getsetall.jl")
 include("sugar.jl")
 include("functionlenses.jl")
 include("testing.jl")
 
-# always included for now
-include("../ext/AccessorsDatesExt.jl")
-include("../ext/AccessorsLinearAlgebraExt.jl")
```
devmotion, line 21: This should be kept (unless you want to demote them to weak dependencies as well in this PR).

.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - 1.6 # LTS
+          - '1.10' # LTS
```
devmotion, line 13:
```suggestion
          - 1.9 # Oldest supported version
          - '1.10' # LTS
```

devmotion:
> the linked julia issue seems close to be fixed.

That doesn't help users on older Julia versions that might run into the issue with circular loading if Accessors is a (possibly quite indirect) dependency and loads Test.

> This is not like an urgent bugfix or even a new feature

It's a fix for users that see these warnings because of Accessors. It also improves loading times for anyone who's not interested in the Test-related functionality. If Accessors keeps Test as a dependency, then it invalidates all the effort that was spent in other packages to not load Test by default since it will trigger all these optional extensions. So Accessors "breaks" the ecosystem.

> And some package in an environment can depend on Test for actual functionality, so it won't be a complete solution anyway

I assume that's not very likely with Test - in very big environments I work with there's not a single package that requires a Test dependency.

devmotion:
> Assuming that it does get fixed on 1.10, there won't be any supported Julia version that shows these warnings

On all older Julia versions (at least 1.9 - 1.10.4) Accessors with a Test dependency can trigger these warnings if it is an indirect dependency of a package that has >= 2 extensions of which one is triggered by Test. The only way to fix this on already released Julia versions is to make Test a weak dependency of Accessors, already on Julia 1.9.

Regardless of these warnings, the following point is also valid on Julia >= 1.9 and not addressed by a new Julia release:

> If Accessors keeps Test as a dependency, then it invalidates all the effort that was spent in other packages to not load Test by default since it will trigger all these optional extensions. So Accessors "breaks" the ecosystem.

devmotion:
I guess there's no point in arguing any further given that you seem not to be open to any of the arguments I've provided so far. I still don't see any advantage in waiting for a new Julia release - by insisting on the Test dependency, Accessors is right now counteracting the effort of the majority of the Julia ecosystem that tries to remove Test as a dependency and Accessors is right now responsible for some of these reports about warnings on Julia >= 1.9 that users see. Both problems could be fixed by this PR and are independent of any new Julia release.

devmotion:
As explained in e.g. [this Julia issue](https://github.com/JuliaLang/julia/issues/52511#issuecomment-1983388150) circular loading can happen whenever
1. You have two or more extensions
2. The triggers for these extensions are in the set of your transitive dependencies.

> imo it is on those packages that made related changes recently, such as making Test a weakdep despite the known Julia issue.

Many packages have made these changes quite a while ago without knowing about potential issues. Now that it's clear what can cause them, the best solution in terms of warnings and loading times (for existing and upcoming Julia versions) is to go through your dependency stack and update other packages as well that have not made Test a weak dependency yet. That's why we updated InverseFunctions and ChangesOfVariables recently, and that's why @ChrisRackauckas opened this PR 🙂

devmotion:
Yes, this seems right:
```julia
(jl_Sjh3WZ) pkg> why Test
  Accessors → Test

(jl_Sjh3WZ) pkg> why Dates
  Accessors → Dates

(jl_Sjh3WZ) pkg> why LinearAlgebra
  Accessors → ConstructionBase → LinearAlgebra
  Accessors → LinearAlgebra
```

As an additional data point: Someone [reported warnings with GeoStatsBase](https://github.com/JuliaMath/InverseFunctions.jl/issues/51#issuecomment-2298549644) after Test was made a weak dependency of InverseFunctions (I could reproduce these warnings). It turned out that CommonSubExpressions, a transitive dependency on GeoStatsBase, depended on Test (even though it was only used for tests). A new release of CommonSubExpressions without the Test dependency was registered today, and with this latest release I can't reproduce the warnings anymore.

nhz2:
I have an alternative PR #167 

I don't like the spooky action at a distance this PR creates where some transitive dependency may add or remove a `using Test` and that changes the behavior of a function in a completely different package.

If these testing functions are part of the API of Accessors then they must reliably work. If they are not part of API of Accessors then they should be removed.

Having functions in a package that sometimes works and sometimes doesn't work is very confusing. For example, if I test my package using Test.jl it will appear to work. Then maybe a few months later, some transitive dependency removes a `using Test`. Still my package passes all tests. Now I go to run a month-long simulation and since `Test` is no longer in the environment the simulation breaks.

devmotion:
> If these testing functions are part of the API of Accessors then they must reliably work.

They work reliably - if you load the package required for the extension.

You could apply your arguments also to any other extension, it seems? The main point of extensions is that they're not always automatically loaded.

nhz2:
No, well design extensions do not behave like this. For example, https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)

```julia
function Plotting.plot(c::Contour.ContourCollection)
    # Some functionality for plotting a contour here
end
```
This will reliably work, because to call this method, `Contour` must have been loaded for the user to have created a `Contour.ContourCollection` object.

In this PR nothing is passed to the testing functions that always ensure `Test` has been loaded.

devmotion:
I think I get your point. But I think such an extension is fine nevertheless if it is documented and users are told they have to load Test for these testing utilities (it's definitely commonly used in other packages). Currently, however, the functions are neither exported nor documented, so I don't think there's anything to worry about.

devmotion:
I'm repeating myself here - but the problem with AccessorsExtras example is not the Julia version but the fact that it has extensions that are triggered by transitive dependencies - this PR will fix the problem for Test, and it's the only stable solution (unless you want to enforce that no dependant of Accessors can make Test a weak dependency).

The ConstructionBase PR is not needed here - IMO LinearAlgebra should be made a weak dependency in a separate release (see above).

devmotion:
Surely base julia could handle the problem better (I'm not familiar with Pkg enough to be of any use there) but even with "nicer" messages or a different precompilation order there's no point in making extensions for anything that's loaded unconditionally anyway. As long as Accessors is keeping Test as a dependency, there's no point for any package that depends on Accessors (such as the AccessorsExtras example) to make Test a weak dependency.

jw3126:
> If others (@jw3126 @rafaqz) are fine with that (minor) risk of these warnings appearing in some complex envs, I guess I'll tag new ConstructionBase and do the Accessors weakdep transition (merge @devmotion's PR & tag) soon.

I am fine either way.

devmotion:
Another set of warnings caused by Accessors that would be resolved if Accessors would make Test a weak dependency (on Julia >= 1.9!): https://github.com/JuliaStats/Distributions.jl/pull/1900#issuecomment-2347904231

devmotion:
There are no issues as long as there are no cycles (which defeats the purpose of an extension anyway - it shouldn't be triggered unconditionally). Adding a dependency on Roots would add such a cycle to Distributions since Distributions has a weak dependency on Test but Accessors, a dependency of Roots, still depends on Test.

> Hmm, so that's not just stdlibs? What's happening there?

It's the same issue I keep telling you about 😄 Distributions has a ChainRulesCore extension but Roots depends on it, so adding Roots as a dependency of Distributions will trigger the ChainRulesCore extension unconditionally which creates a circular dependency which Pkg warns about.

rafaqz:
Since I was pinged earlier, I think we need to both get rid of direct dependencies on e.g. Test, and at the same time any circular dependencies in the package ecosystem.

This PR is just one of those steps. I don't see any problem with this, and agree 100% with @devmotion

rafaqz:
Sorry, I don't get this hypothetical. What specific packages adding dependencies can cause circular dependency problems for Accessors extensions? Isn't this a very small set, so this problem can be solved socially? 

Just the people commenting in this PR very likely have the social capital to get basically any needed PRs merged.

Further, I'm personally maintaining quite a few packages with up to five extensions, and don't experience these problems.

devmotion:
> but we had to revert them.

That's just wrong - the warnings in https://github.com/JuliaObjects/Accessors.jl/issues/127 were again caused by the same problem. You could have fixed them by going through the list of dependencies and ensuring that none of them (directly or indirectly) depends on any of the weak dependencies of Accessors. From the Manifest file in that issue it is clear that you had made LinearAlgebra a weak dependency of Accessors even though ConstructionBase (dependency of Accessors) depended on LinearAlgebra and you had made Test a weak dependency even though InverseFunctions (dependency of Accessors) depended on Test.

rafaqz:
Can you link the line where there is a circular dependency break? I can see it

rafaqz:
ConstructionBase.jl is still causing at least one of those, why don't we just start fixing the ones we know that @devmotion already said we need to fix

rafaqz:
Looks like LinearAlgebra is still in `deps` to me ;)

longemen3000:
On Julia 1.9 onwards, pkg ignores entries that are in [deps] and [weakdeps]. This is to maintain backward compatibility.

rafaqz:
Ok sorry for the noise, but then is the problem not just going through all the Accessors dependencies to analyze their direct deps? Has that been done?

devmotion:
The actual Manifest file that lists the chain of dependencies would be much more helpful than the logs of the CI run. If I checked correctly, none of the remaining dependencies of Accessors depends on LinearAlgebra anymore (at least not in the latest versions - but many users can't update to the latest ConstructionBase version, including myself, which is why I'm advocating for doing the LinearAlgebra change in a separate PR - if at all). It's pure speculation but I assume that the problem is that some of the weak dependencies (such as [AxisKeys](https://github.com/mcabbott/AxisKeys.jl/blob/master/Project.toml)) actually depend on LinearAlgebra. I imagine this can also cause cycles when trying to load these weak dependencies since AxisKeys will trigger the weak extension which loads both AxisKeys and Accessors but at the same time it will also trigger (through the indirect dependency on LinearAlgebra) the weak extension for LinearAlgebra which itself then tries to load both LinearAlgebra and Accessors.

devmotion:
Maybe one could help Pkg to figure out the desired loading order by including LinearAlgebra in the list of triggers for the weak extension for AxisKeys?

rafaqz:
Right so at least a LinearAlgabraExt PR to AxisKeys.jl is needed. Then probably others.

It is showing that having extensions on mid level packages like AxisKeys in a low level packages like Accessors leads to the low level one owning some of the others maintenance work

devmotion:
>  Even when given a concrete situation, it's not clear what the underlying reason for these warnings is.

It's speculation because did not have access to the Manifest file and did not have time to investigate it in more detail myself.

>  It would be somewhat more comfortable if anyone who developed/develops Julia extension machinery confirmed that some extensions (Dates? Test?) are safe to add this situation.

There's a single problem: cycles. So no package/extension is safe per se, you just have to check (or be told by Pkg) that you do (not) create a cycle by moving something to an extension.

> It's really brittle as your comments indicate. For example, Unitful depends on Dates, maybe it causes the warnings?

As pointed out previously, I think it's a design flaw to have these extensions in Accessors. IMO Accessors should have only few dependencies and also only few weak dependencies, given that it's not a leaf package but rather used by many packages in the ecosystem. My feeling is also that such dependency cycles would be less likely if e.g. the Unitful extension would be an Accessors extension in Unitful instead (and similarly e.g. the AxisKeys extension).

rafaqz:
@devmotion I agree on limiting extensions here, except for very low level packages like Dates and LinearAlgebra. With Unitful Accessors could just limit itself to using the ConstructionBase compatibility I added to Unitful.jl

PR #166, opened 2024-08-25
improve error message

aplavin:
Was a bit confusing, eg:
```julia
julia> set(1, year, 123)
ERROR: This should be unreachable. You probably need to overload
`Accessors.set(obj, ::typeof(year), val)
```
even though `set()` with `year` already exists, but defined for dates only.
With this PR:
```julia
julia> set(1, year, 123)
ERROR: This should be unreachable. You probably need to overload
`Accessors.set(obj::Int64, ::typeof(year), val::Int64)`
```


jw3126 (approved):
Thanks

PR #168, opened 2024-08-29
Make Dates and Tests weak dependencies

devmotion:
LinearAlgebra is kept as a direct dependency until https://github.com/JuliaObjects/ConstructionBase.jl/pull/95 is released: Without that PR, LinearAlgebra is a transitive dependency of Accessors and hence a LinearAlgebra extension of Accessors is not useful (and will even lead to warnings).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Accessors"
 uuid = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 authors = ["Takafumi Arakaki <aka.tkf@gmail.com>", "Jan Weidner <jw3126@gmail.com> and contributors"]
-version = "0.1.37"
+version = "0.1.38"
 
 [deps]
 CompositionsBase = "a33af91c-f02d-484b-be07-31d278c5ca2b"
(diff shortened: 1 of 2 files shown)


devmotion:
> This PR indeed doesn't seem to cause issues within Accessors tests.

That's good, it's supposed to be a fix 🙂

> But I tried to enable all weakdeps (this PR + linearalgebra), and then one gets that "big scary warning" even when running ]test for Accessors.

I can't say much without knowing the packages and their versions that were used when you ran `Pkg.test`. But I'm very certain that one of the (test) dependencies depends on LinearAlgebra, which then again created a cycle. Every problem and warning I've encountered so far is caused by this problem 🙂 Maybe you tested with an older version of ConstructionBase?

devmotion:
> Having an extra dependency is not a bug, so removing it cannot be a "fix". It's strictly an optimization, and a pretty minor one for stdlibs.

Currently, Accessors breaks/holds back large parts of the ecosystem and this is resolved by the PR. I'd consider that a fix.

devmotion:
Basically every package tries to move `Test` utilities to an extension and considers a dependency on `Test` a bug. But downstream packages that depend on Accessors can't fix these issues as long as Accessors keeps enforcing an indirect dependency on `Test`.

PR #169, opened 2024-08-30
Update GH actions + add dependabot

devmotion:
The PR updates the GH actions, adds dependabot (to keep them updated automatically), ~~and uses `lts` instead of 1.6 as a more stable way to test LTS versions~~.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - lts
+          - 1.6 # LTS
           - 1
           - 'pre'
           - 'nightly'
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -10,7 +10,7 @@ jobs:
       fail-fast: false
       matrix:
         version:
-          - 1.6 # LTS
+          - lts
```
jw3126, line 13: We can't require `lts` in project.toml, right? So I think an explicit version in CI might be better.

jw3126 (approved):

.github/workflows/ci.yml
devmotion: That's a good point. Ideally, I guess one would want to test the oldest supported version and LTS - the first could be hardcoded (since one has to update the Project.toml manually anyway when changing it) and the second achieved with the `lts` option. But it's unnecessarily many tests if both coincide...

PR #175, opened 2024-10-08
tests: add workaround for regression in 1.11

aplavin:
Good news: 1.11 doesn't really break anything in Accessors, not even any `@inferred` test :)

The only (surprisingly, still unresolved) issue affecting us is https://github.com/JuliaLang/julia/issues/54664. There's already a PR https://github.com/JuliaObjects/Accessors.jl/pull/170 to work it around, but that one uses internals. I asked on slack today and got a recommendation to just load `REPL` – it works indeed!


jw3126:
My hope would be that it gets fixed upsteam, so I could live with merging and this test failing for a while. If it never gets fixed upsteam and no brave soul wants to look into it, we can remove the test.

jw3126:
you want to bump the patch version?

ranocha:
Yeah, it looks like it's an upstream error in JET.jl: https://github.com/JuliaObjects/Accessors.jl/actions/runs/11242478221/job/31256238698?pr=175#step:5:449
Did you report it there? From my point of view, going ahead with this PR even when the upstream error persists is fine.

PR #190, opened 2025-01-07
use shorter extension names

aplavin:
The shorter names have been recommended for quite some time already, see https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions). Actually, there was only a short period of time early during 1.9 version, when some tooling didn't work with these shorter names. Now, they only serve to clutter precompilation output :)

### DIFF
--- a/src/Accessors.jl
+++ b/src/Accessors.jl
@@ -17,14 +17,14 @@ include("functionlenses.jl")
 include("testing.jl")
 
 if !isdefined(Base, :get_extension)
-    include("../ext/AccessorsDatesExt.jl")
-    include("../ext/AccessorsLinearAlgebraExt.jl")
-    include("../ext/AccessorsTestExt.jl")
+    include("../ext/DatesExt.jl")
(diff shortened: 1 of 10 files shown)


jw3126 (approved):
LGTM

src/Accessors.jl
```diff
@@ -17,14 +17,14 @@ include("functionlenses.jl")
 include("testing.jl")
 
 if !isdefined(Base, :get_extension)
-    include("../ext/AccessorsDatesExt.jl")
-    include("../ext/AccessorsLinearAlgebraExt.jl")
-    include("../ext/AccessorsTestExt.jl")
+    include("../ext/DatesExt.jl")
+    include("../ext/LinearAlgebraExt.jl")
+    include("../ext/TestExt.jl")
 end
```
jw3126, line 23: Not this PR. But what do you think about bumping julia version to the current LTS 1.10 and getting rid of Requires and other legacy compat?

PR #191, opened 2025-01-08
CI settings

aplavin:
- wasn't tested on macos for some reason – added
- tests ran on nightly, leading to failures often reported; in practice, we don't even try to fix those issues, and now that the `pre` version exists – let's just remove `nightly`
- run the Documentation workflow on 1.10

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -38,7 +38,7 @@ jobs:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
         with:
-          version: 1.7
+          version: '1.10'
       - run: |
           julia --project=docs -e '
             using Pkg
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -14,10 +14,9 @@ jobs:
...
           - '1'
           - 'pre'
-          - 'nightly'
         os:
           - ubuntu-latest
-          # - macOS-latest
+          - macOS-latest
           - windows-latest
         arch:
           - x64
```
devmotion, line 22: You should probably remove `arch` completely and just rely on the default of setup-julia (uses the architecture of the runner): macOS-latest runs Apple Silicon, so with `x64` you won't run the matching Julia binary.
```suggestion
```

PR #193, opened 2025-01-10
revert some weakdeps, fix #192

### LINKED ISSUE #192: Circular dependency with Accessors 0.1.40 and julia 1.10
When installing the latest Accessors 0.1.40 on julia 1.10 I get a circular dependency warning

```julia
┌ Warning: Circular dependency detected. Precompilation will be skipped for:
│   DatesExt [0361c7f5-3687-5641-8bd2-a1de0c64d1ed]
│   LinearAlgebraExt [ef8e1453-9c17-56fe-886b-405471570bc8]
```

It gives no warning on julia 1.11 or installing Accessors 0.1.39 (whose project is a bit different but not substantially so from 0.1.40). Maybe it is an upstream bug in julia 1.10?

aplavin:
fix #192

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,7 +1,7 @@
 name = "Accessors"
 uuid = "7d9f7c33-5ae7-4f3b-8dc6-eff91059b697"
 authors = ["Takafumi Arakaki <aka.tkf@gmail.com>", "Jan Weidner <jw3126@gmail.com> and contributors"]
-version = "0.1.40"
+version = "0.1.41"
 
 [deps]
 CompositionsBase = "a33af91c-f02d-484b-be07-31d278c5ca2b"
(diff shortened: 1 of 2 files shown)


Project.toml
```diff
@@ -15,9 +15,7 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 
 [weakdeps]
 AxisKeys = "94b1ba4f-4ee9-5380-92f1-94cde586c3c5"
-Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
 IntervalSets = "8197267c-284f-5f27-9208-e0e47529a953"
-LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
```
devmotion, line 20: AFAICT to break the cycle due to the loading behaviour on Julia 1.10 it would be sufficient to make one of Dates and LinearAlgebra a regular dependency. This is also how it was in the previous release it seems (only LinearAlgebra was added as a weak dependency). I would guess that LinearAlgebra triggers more and slower to load extensions in other packages than Dates, so it might be beneficial to keep LinearAlgebra as a weak dependency and only make Dates a regular dependency.

ChrisRackauckas (approved):

PR #195, opened 2025-02-04
tweak NamedTuple operations to be more type-stable

aplavin:
Don't know how to make a self-contained test for it, but found that this change improves stability in some cases.
I guess Julia prefers if types stay in the type domain the whole time (as in `NamedTuple{(field,)}`), instead of going through the value domain and constprop (as in `(; field => val)`).

### DIFF
--- a/src/optics.jl
+++ b/src/optics.jl
@@ -389,6 +389,7 @@ function (l::PropertyLens{field})(obj) where {field}
     getproperty(obj, field)
 end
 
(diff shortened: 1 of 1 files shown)


jw3126 (approved):
LGTM, maybe add a comment so that nobody cleans up this "ugly" syntax accidentally.

jw3126 (approved):

PR #196, opened 2025-02-18
Allow for proposed new parsing of `array[end]` in Accessors

### LINKED ISSUE #57269: `arr[var"end"]` is the same as `arr[end]`
Brought up in #56936 (thanks @martinholters for the example):

```julia
julia> let var"end"=1; (1:10)[var"end"]; end
10 # anything else than "end", and you get 1
```

mlechu:
To fix https://github.com/JuliaLang/julia/issues/57269, a small-but-breaking AST change is needed (PR https://github.com/JuliaLang/julia/pull/57368).  The `set` macro in Setfield and Accessors breaks, and enough packages depend on this macro that it needs to be updated before deciding whether to merge the julialang PR.

This PR only recognizes the new way of parsing `array[end]`, and should not change any behaviour when used with the current parser.

### DIFF
--- a/src/sugar.jl
+++ b/src/sugar.jl
@@ -141,7 +141,7 @@ foldtree(op, init, ex::Expr) =
 
 need_dynamic_optic(ex) =
     foldtree(false, ex) do yes, x
-        yes || x === :end || x === :begin || x == Expr(:end) || x == Expr(:begin) || x === :_
+        yes || x ∈ (:end, :begin, Expr(:end), Expr(:begin), :_)
     end
 
 replace_underscore(ex, to) = postwalk(x -> x === :_ ? to : x, ex)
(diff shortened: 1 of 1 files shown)


mlechu:
Thanks!

> I assume that if the linked Julia PR isn't accepted in the end, we should remove this code? Also, not forget to modify it if that PR evolves on the way.

Yes, I'll update/undo these changes if they become obsolete.

aplavin (approved):

jw3126:
Just for reference it is very likely that this AST change will happen in julia and @keno recommends merging this PR: https://github.com/JuliaLang/julia/pull/57368#issuecomment-2667272476

mlechu:
@jw3126 Thanks!  Whenever you have time, could you add this change to a release for both packages so that I can check how the parsing change affects the packages depending on setfield and accessors?

PR #202, opened 2025-06-08
remove obsolete filter from CI

aplavin:
... also check whether tests fail on the upcoming julia version

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -17,9 +17,6 @@ jobs:
           - ubuntu-latest
           - macOS-latest
           - windows-latest
-        exclude:
-          - os: macOS-latest # Apple Silicon
-            version: '1.6'
     steps:
       - uses: actions/checkout@v4
       - uses: julia-actions/setup-julia@v2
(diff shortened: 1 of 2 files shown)


aplavin:
Indeed, https://github.com/JuliaObjects/Accessors.jl/blob/bd9151366236a4b012683619d430b81105deb057/test/perf.jl#L181 fails (composition order).
It appears to be a genuine regression in Julia. But also not sure what's the relationship of this test to Accessors, why L vs R composition orders are compared in the tests here... @jw3126 do you remember?

jw3126:
A long time ago in julia a composed lens was fast or slow depending on the composition order. E.g. these two had different performance:
```julia
@optic(_.d) ∘ (@optic(_.c) ∘ (@optic(_.b) ∘ @optic(_.a)))
((@optic(_.d) ∘ @optic(_.c)) ∘ @optic(_.b)) ∘ @optic(_.a)
```
The purpose of the test was to check that `@optic _.d.c.b.a` gives us the fast one. Then julia improved and both orders became fast. Now it seems there is a regression and we again have a fast and a slow order.
Probably we should change the test to be:
`@test min(b_left.time, b_right.time) ≈ b_default.time rtol=0.8 # no matter which order is faster, it is the default`
E.g. we only care to get the fast one, but we don't really care in the test which one is the fast one.

aplavin:
Thanks for the background, updated the test accordingly and added a comment.

PR #205, opened 2025-08-02
even friendlier compact show

aplavin:
Doesn't change the default `show` – it remains parseable, same as before.
Tweaks compact show: it already wasn't Julia-parseable, here I make its output even friendlier. See examples in the tests! Mechanism is reasonably extensible, with the internal `_shortstring` function receiving `is_compact` argument.


aplavin:
after thinking more about this – too much special casing for default behavior, imo
functionality is available in AccessorsExtra – use its `sprint(AccessorsExtra.show_optic, youroptic)`

PR #208, opened 2025-08-05
Add PropertyLensRuntime

### LINKED ISSUE #145: String properties not supported
```julia
julia> @optic _."abc"
ERROR: TypeError: in Type, in parameter, expected Type, got a value of type String
```
This is because PropertyLens is defined to hold the property name as a type parameter, and Strings cannot be type params.
Not sure what we can do, but maybe someone has ideas on how to fix it?

aplavin:
Fixes #145

Adds PropertyLensRuntime.
Also, switches to ComposedFunction instead of ComposedOptic everywhere.

### DIFF
--- a/src/optics.jl
+++ b/src/optics.jl
@@ -1,4 +1,4 @@
-export PropertyLens, IndexLens
+export PropertyLens, PropertyLensRuntime, IndexLens
 export set, modify, delete, insert, getall, setall
 export ∘, opcompose, var"⨟"
 export Elements, Recursive, If, Properties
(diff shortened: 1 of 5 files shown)


src/optics.jl
```diff
@@ -398,6 +400,21 @@ end
...
 @inline insert(obj::Tuple, l::PropertyLens{field}, val) where {field} = insert(obj, IndexLens(field), val)
 
+"""
+    PropertyLensRuntime(propertyname)
+
+Construct a lens for accessing a property `propertyname` of an object where the property name
+is stored as a runtime value rather than a type parameter. This is useful when the property
+name cannot be a type parameter (e.g., strings, arbitrary objects).
+
+Automatically supports getting a value, but `set` requires explicit support for each object type.
```
jw3126, line 410: I wonder if we should supply a default implementation using setproperties?

jw3126 (approved):

src/optics.jl
aplavin: `setproperties` only supports Symbol and Ints, because it takes NamedTuple or Tuple only – for type stability. In principle ConstructionBase can be extended to support more through Dicts. It would be a more general approach indeed, even though I don't see how a default implementation of `setproperties` is possible for this scenario.

Anyway, string properties don't come up that often though, so I'm personally fine with just supporting them here in Accessors – not in CB.jl.

PR #213, opened 2025-09-13
`set(arr, reverse, arr2)` can change size

### LINKED ISSUE #212: DimensionMismatch after slicing
I was expecting this to work.

```jl
julia> let obj  = ['h', 'e', 'l', 'l', 'o']
           @modify(reverse(obj)) do x
               x[1:3]
           end
       
       end
ERROR: DimensionMismatch: array could not be broadcast to match destination
Stacktrace:
 [1] check_broadcast_shape
   @ ./broadcast.jl:552 [inlined]
 [2] check_broadcast_axes
   @ ./broadcast.jl:555 [inlined]
 [3] instantiate
   @ ./broadcast.jl:310 [inlined]
 [4] materialize!
   @ ./broadcast.jl:883 [inlined]
 [5] materialize!
   @ ./broadcast.jl:880 [inlined]
 [6] set(x::Vector{Char}, ::typeof(reverse), v::Vector{Char})
   @ Accessors ~/.julia/packages/Accessors/8Y4uq/src/functionlenses.jl:96
 [7] _modify(f::var"#153#154", obj::Vector{Char}, optic::typeof(reverse), ::Accessors.SetBased)
   @ Accessors ~/.julia/packages/Accessors/8Y4uq/src/optics.jl:209
 [8] modify(f::Function, obj::Vector{Char}, optic::typeof(reverse))
   @ Accessors ~/.julia/packages/Accessors/8Y4uq/src/optics.jl:189
 [9] top-level scope
```

jariji:
Allow `reverse`d objects to change size. E.g. 

```jl
julia> @modify([10,20,30]) do xs
           xs[2:3]
       end
2-element Vector{Int64}:
 20
 30
```

Resolves #212.

### DIFF
--- a/ext/AxisKeysExt.jl
+++ b/ext/AxisKeysExt.jl
@@ -10,4 +10,11 @@ Accessors.set(x::KeyedArray, f::Base.Fix2{typeof(axiskeys), Int}, v) = @set axis
 Accessors.set(x::KeyedArray, f::Base.Fix2{typeof(axiskeys), Symbol}, v) = @set named_axiskeys(x)[f.x] = v
 
 Accessors.set(x::KeyedArray, ::typeof(AxisKeys.keyless_unname), v::AbstractArray) = KeyedArray(v; named_axiskeys(x)...)
+
+function set(x::AbstractVector, ::typeof(reverse), v::KeyedArray)
+    res = similar(x, eltype(v))
+    res .= v
+    reverse!(res)
+    res
+end
 end
--- a/ext/AxisKeysExt.jl
+++ b/ext/AxisKeysExt.jl
@@ -11,7 +11,7 @@ Accessors.set(x::KeyedArray, f::Base.Fix2{typeof(axiskeys), Symbol}, v) = @set n
 
 Accessors.set(x::KeyedArray, ::typeof(AxisKeys.keyless_unname), v::AbstractArray) = KeyedArray(v; named_axiskeys(x)...)
 
-function set(x::AbstractVector, ::typeof(reverse), v::KeyedArray)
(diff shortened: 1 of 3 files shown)


jariji (commented):
I'm not sure why `set` isn't hitting the method in the extension. If I define those methods directly in the repl the right one is selected and the test passes.

```jl
julia> using AxisKeys,Accessors, Test

julia> function Accessors.set(x::AbstractVector, ::typeof(reverse), v)
           res = similar(x, eltype(v))
           res .= v
           reverse!(res)
           res
       end
       function Accessors.set(x::AbstractVector, ::typeof(reverse), v::AbstractVector)
           res = similar(x, eltype(v), size(v))
           res .= v
           reverse!(res)
           res
       end
       function Accessors.set(x::AxisKeys.KeyedArray, ::typeof(reverse), v::AbstractVector)
           res = similar(x, eltype(v))
           res .= v
           reverse!(res)
           res
       end
       let
           C = KeyedArray([1,2,3], x=[:a, :b, :c])
           @test (@set reverse(C) = [5,6,7])::KeyedArray == KeyedArray([7,6,5], x=[:a, :b, :c])
       end
Test Passed
```

aplavin:
Thanks for trying it out @jariji!

It's not just KeyedArrays that need special handling after this PR. For example, StaticArrays now lose their "staticness" completely: `set(SVector(1,2,3), reverse, 10:12)` now returns a plain Vector without a static length.
Maybe there are even more cases like that... But even with 2 major "special cases", I'm personally not fully convinced this is the way to go.

aplavin:
But the extension issue itself is simple: you define a new `set` function that doesn't overload `Accessors.set`.

PR #217, opened 2025-10-17
Implement setindex for Base.ImmutableDict

### LINKED ISSUE #216: Add support for Base.ImmutableDict
I tried to use Accessors with `Base.ImmutableDict`:

```julia
julia> x = Base.ImmutableDict("a" => 1, "b" => 2)
Base.ImmutableDict{String, Int64} with 2 entries:
  "b" => 2
  "a" => 1

julia> @set x["b"] = 3
ERROR: MethodError: no method matching empty!(::Base.ImmutableDict{String, Int64})
The function `empty!` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  empty!(::ComputePipeline.ComputeGraph)
   @ ComputePipeline ~/.julia/packages/ComputePipeline/03tW7/src/ComputePipeline.jl:1041
  empty!(::Scene; reset_theme)
   @ Makie ~/.julia/packages/Makie/4JW9B/src/scenes.jl:471
  empty!(::Compiler.CompilationQueue)
   @ Base ~/.julia/juliaup/julia-1.12.0+0.aarch64.apple.darwin14/share/julia/Compiler/src/typeinfer.jl:1333
  ...

Stacktrace:
 [1] copy!(dst::Base.ImmutableDict{String, Int64}, src::Base.ImmutableDict{String, Int64})
   @ Base ./abstractdict.jl:200
 [2] setindex
   @ ~/.julia/packages/Accessors/8Y4uq/src/setindex.jl:22 [inlined]
 [3] set(obj::Base.ImmutableDict{String, Int64}, lens::IndexLens{Tuple{String}}, val::Int64)
   @ Accessors ~/.julia/packages/Accessors/8Y4uq/src/optics.jl:417
 [4] top-level scope
   @ REPL[62]:1
```

It would be nice to add support for `Base.ImmutableDict`.

danielmatz:
The design of Base.ImmutableDict allows the same key to be added again, which shadows the previously added value. Unfortunately, it still prints out all of the entries, including repeated keys. I thought this might be confusing, so I implemented it to remove the value if the key exists and then add the new value.

Closes #216

### DIFF
--- a/src/setindex.jl
+++ b/src/setindex.jl
@@ -24,6 +24,10 @@ Base.@propagate_inbounds function setindex(d0::AbstractDict, v, k)
     return d
 end
 
+function setindex(d::Base.ImmutableDict, v, k)
+    Base.ImmutableDict(d, k => v)
+end
+
 # may be added to Base in https://github.com/JuliaLang/julia/pull/46453
 @inline function setindex(t::Tuple, v, inds::AbstractVector{<:Integer})
     ntuple(length(t)) do i
--- a/test/test_setindex.jl
+++ b/test/test_setindex.jl
@@ -43,6 +43,10 @@ end
     @test Accessors.setindex(d, 30, "c") ==ₜ Dict(:a=>1, :b=>2, "c"=>30)
     @test Accessors.setindex(d, 10.0, :a) ==ₜ Dict(:a=>10.0, :b=>2.0)
 
+    d = Base.ImmutableDict(:a => 1, :b => 2)
+    @test @set(d[:a] = 3) == Base.ImmutableDict(:a => 1, :b => 2, :a => 3)
+    @test @set(d[:c] = 3) == Base.ImmutableDict(:a => 1, :b => 2, :c => 3)
+
     nt = (a=1, b='2')
     @test @set(nt[:a] = "abc") == (a="abc", b='2')
     @test @set(nt[1] = "abc") == (a="abc", b='2')


src/setindex.jl
```diff
@@ -24,6 +24,16 @@ Base.@propagate_inbounds function setindex(d0::AbstractDict, v, k)
...
 end
 
+function setindex(d0::Base.ImmutableDict, v, k)
+    d = Base.ImmutableDict(
+        (
+            key => value
+            for (key, value) in d0 if key != k
+        )...
+    )
+    Base.ImmutableDict(d, k => v)
```
aplavin:
```suggestion
    d = Base.ImmutableDict(
        (
            key => value
            for (key, value) in d0 if key != k
        )...,
        k => v
    )
```
?

danielmatz:
I'm going to wait for the conversation on my PR to implement `Base.setindex` to play out, since that will likely influence how we should implement this method here.

danielmatz:
Based on the discussion about `Base.setindex`, it seems that `ImmutableDict` was intended to behave like a multi-dict. So, existing keys should remain. I updated this branch to not remove existing duplicate keys.

cstjean (approved):
Uneasy about that one, but it's the Base behaviour...

aplavin:
I suggest we wait for https://github.com/JuliaLang/julia/pull/59880 to be approved, to make sure we won't need to change the semantics afterwards.
Unless it's urgent for @danielmatz?

danielmatz:
This isn't urgent for me at all.

PR #218, opened 2025-12-01
performance optimizations

aplavin:
`Accessors` are known to be hard for the compiler to optimize sometimes, with compiler heuristics to stop optimizing triggering quite early.
Here, I make the compiler "try harder": things like avoiding argument splatting, and making small inner functions more optimizable.

I recently used Accessors getall/setall with a few tens of elements – up to ~50. With these optimizations, it compiles down to almost nothing, while before there was a lot of dynamic dispatch inside.
Tests pass both before and after. Maybe I'll think of a short self-contained test demonstrating the improvements and add it... But I confirm in my testing that these changes greatly improve performance in such a regime.

### DIFF
--- a/src/getsetall.jl
+++ b/src/getsetall.jl
@@ -113,7 +113,7 @@ _getall(obj, optics::Tuple{Any}) = getall(obj, only(optics))
 for N in [2:10; :(<: Any)]
     @eval function _getall(obj, optics::NTuple{$N,Any})
         _reduce_concat(
-            map(getall(obj, last(optics))) do obj
+            mapspec(getall(obj, last(optics))) do obj
                 _getall(obj, Base.front(optics))
             end
         )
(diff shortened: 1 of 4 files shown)


src/getsetall.jl
```diff
@@ -134,11 +134,18 @@ end
 
 # helper functions
 
-_concat(a::Tuple, b::Tuple) = (a..., b...)
+@generated _concat(a::NTuple{N,Any}, b::NTuple{M,Any}) where {N,M} =
+	:(Base.Cartesian.@ntuple $(N + M) i -> i ≤ $N ? a[i] : b[i - $N])
```
jw3126, line 140: There is no reliably fast way to do this in Base?

jw3126 (approved):

src/getsetall.jl
aplavin: I tried:
- `map`
- broadcast
- `ntuple` the function

All of these suffer from catastrophic performance drop at some point. _Seems like_ it's around 32 elements returned by `getall` in total, the unrolling threshold for tuples in Base.

Added a comment.

aplavin:
Ideally, Accessors should work and be (almost) zero cost in all regimes – from heterogeneous tuples at small sizes, to homogeneous abstractvectors after some threshold. Unfortunately, the seamless transition is notoriously difficult, don't think we ever seriously attempted it.
What we can do cheaply (this PR) is to expand the "tuples" regime to larger sizes – there are still fundamental limits, but multiple tens of elements works great now.

PR #220, opened 2026-03-13
Use eltype(v) instead of promote_type when replacing all array elements

aplavin:
## Summary
- Add `_setindex_fresh` helper that allocates with `eltype(v)` directly (no `promote_type`, no `copy!`) for cases where all elements are being replaced
- Use it for Colon `setindex`, `sort` (set+modify), and `vec`
- Fixes e.g. `@set sort([1,2,3]) = ["a","b","c"]` returning `Vector{Any}` instead of `Vector{String}`

## Test plan
- Added `@inferred` and `==ₜ` tests for Colon and sort with different eltypes
- Full test suite passes


aplavin:
I'll go ahead and merge, given that it's a pretty focused self-contained change

